Issue #1, opened 2021-04-26
Break out of dev/XAI the saliency family of interfaces into a new "xaitk-saliency" package here.


Purg:
Kitware/SMQTK branch "dev/XAI".
Sources: https://github.com/Kitware/SMQTK/tree/dev/XAI/python/smqtk/algorithms/saliency

- [x] Break out files using git-filter-repo to maintain history context
- [x] Adopt support files and configs from smqtk-core.
- [x] Add github actions workflow for unittests and checks
- [x] Restructure files into new package structure.
- [x] Initial pass at adding/fixing type annotations and consistency

Aiming to complete this by May 03.

PR #2, opened 2021-04-27
Add initial structural files

Purg:
Modeled after current SMQTK-* package tooling.


Purg:
Addressing the restructure part of #1

PR #3, opened 2021-04-28
Initial pass at fixing flake8 issues and adding type annotations

Purg:
Attempting an initial pass at resolving mypy issues as they crop up,
though I do not expect to clear them all right now due to the initial
pass qualification.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -21,9 +21,30 @@ disallow_untyped_defs = True
 
 ###############################################################################
 # TPL-module options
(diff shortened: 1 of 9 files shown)


Purg:
@vbhavank Here is where I'm making an initial pass at type annotation and attachment to the current SMQTK-* modules as base dependencies.

Purg:
@vbhavank This is good to go now for the purpose of this branch.

Purg:
Unit tests are expected to fail because there are no unit tests (no tests == failure).

PR #5, opened 2021-05-12
Draft interfaces as Code

Purg:
These are new interfaces as proposed in our recently pushed out API draft. Included are perturbation generation, and visual saliency map generation from similarity and classifier blackboxes.


Purg:
I realize that there is still a "TODO" in a doc-string that I need to fill out. I think the remainder of what's happening here could be reviewed in the mean time.

vbhavank (approved):

xaitk_saliency/interfaces/perturb_image.py
```diff
@@ -0,0 +1,53 @@
...
+
+        Output mask matrices should be congruent in length to the number of
+        perturbed images output, as well as share the same height and width
+        dimensions.
+        These masks should indicate the regions in the corresponding perturbed
+        image that has been modified.
+        Values should be in the [0, 1] range, where a value closer to 1.0
+        indicate areas of the image that are *unperturbed*.
+        Mask values should be of a floating-point type in order to allow for
+        fractional perturbation.
```
brianhhu, line 38: Just a quick thought- when using masks, having them be binary can be pretty useful (i.e. less memory and less storage if cached). Do we plan to support binary masks or should they always be float?

xaitk_saliency/interfaces/vis_sal_classifier.py
```diff
@@ -0,0 +1,84 @@
...
+    For classifiers that generate many class label predictions, it is intended
+    that only a subset of relevant class predictions need be provided here if
+    computational performance is a consideration.
+    """
+
+    @abc.abstractmethod
+    def generate(
+        self,
+        image_conf: np.ndarray,
+        perturbed_conf: np.ndarray,
```
brianhhu, line 33: Just to check my understanding- the implementation is responsible for checking that the selected class indices in image_conf and perturbed_conf align?

xaitk_saliency/interfaces/vis_sal_detection.py
```diff
@@ -0,0 +1,75 @@
...
+        perturb_masks: np.ndarray,
+    ) -> np.ndarray:
+        """
+        Generate visual saliency heat-map matrices for each reference
+        detection, describing what visual information contributed to the
+        associated reference detection.
+
+        We expect input detections to come from a black-box source that outputs
+        our minimum requirements of a bounding-box, objectness and per-class
+        scores.
```
brianhhu, line 39: Maybe add something like: "For architectures which do not produce objectness scores, a default value of 1.0 is used." This goes back to the question of how we plan to encapsulate a detection object.

brianhhu (commented):
Overall, it looks good to me. I left a few minor comments in parts of the code.

xaitk_saliency/interfaces/perturb_image.py
Purg: That's a good point. Maybe a better way to phrase this is that masks "may be float-valued in order allow for fractional perturbation." The important bit is the value range of [0,1] I think. As long as that promise is made/adhered-to, then maybe the value representation is not important. We could show in examples that where appropriate, numpy/tensor type casting should be performed (which is generally a no-op if the input is already the type requested).

xaitk_saliency/interfaces/vis_sal_classifier.py
Purg: I want to say that we want to convey that implementors should be able to assume they are the same dimensionality. We could employ the template method here to inject specific dimensionality assertions to make that check automatic and explicit.

xaitk_saliency/interfaces/vis_sal_detection.py
Purg: > This goes back to the question of how we plan to encapsulate a detection object.

Yeah, I think we have a number of options on that front. My thought there was to use some basic types in this first pass (i.e. numpy), and then with some additional time maybe we could define a "DetectedObjectSet" (maybe defined in smqtk-detectors) that exposes an access API for the required bits, enforcing acceptable defaults where applicable. I think that's a few steps ahead of where we currently are, though.

Purg: I like your recommended phrasing, maybe I'll tweak that though to say that the objectness column should still be provided, but provide the guidance of using the 1.0 value in the absence of it being a natural output of some detector impl.

brianhhu: Sounds good to me!

xaitk_saliency/interfaces/perturb_image.py
brianhhu: Sounds good to me!

xaitk_saliency/interfaces/vis_sal_classifier.py
brianhhu: Resolved based on offline conversations.

Purg:
@brianhhu I amended the appropriate commits to take into account your comments. Thanks!

nitesh201 (approved):

brianhhu (approved):

PR #6, opened 2021-05-16
Occlusion Based (Fsal) implementation of ImageClassifierSaliencyMapGenerator 

vbhavank:
This PR contains example usage of an occlusion based saliency algorithm that makes use of the following saliency API's:
1) ImageClassifierSaliencyMapGenerator
2) PerturbImage

The above API is leveraged to implement a sliding window perturbed along with a classifier-based saliency generation scheme.


vbhavank:
@Purg @brianhhu @nitesh201 This is a PR for pulling in Fsal implementation using the new ImageClassifierSaliencyMapGenerator API. The following branch also contains examples showing usage of the new implementation. Feel free to drop in comments or corrections when you get a chance. Thank you!

xaitk_saliency/impls/saliency/fsal.py
```diff
@@ -0,0 +1,65 @@
+from xaitk_saliency import ImageClassifierSaliencyMapGenerator
+from xaitk_saliency.utils.masking import weight_regions_by_scalar
+
+import numpy as np
+
+
+class Fsal (ImageClassifierSaliencyMapGenerator):
```
brianhhu, line 7: I'm not sure I would call this implementation Fsal, since this is pretty Kitware-specific. It essentially is an occlusion-based saliency algorithm, so maybe something like Occlusion instead?

xaitk_saliency/impls/saliency/fsal.py
```diff
@@ -0,0 +1,65 @@
...
+
+        # Creating an emtpy array of length nClasses and shape H X W
+        sal = np.empty((len(image_conf), *perturbed_masks.shape[-2:]))
+
+        # Iterating through each class confidence and compare it with
+        # its perturbed twin
+        for i, base_class_conf in enumerate(image_conf):
+            diff = np.empty(len(perturbed_conf[:, i]))
+            for j in range(len(perturbed_conf[:, i])):
+                diff[j] = base_class_conf - perturbed_conf[j][i]
```
brianhhu, line 57: I feel like this can be vectorized (maybe with some broadcasting)? Looping over each class does not seem to be the most efficient way to do this.

xaitk_saliency/impls/saliency/sliding_window.py
```diff
@@ -0,0 +1,120 @@
...
+        :param image_size: the mask size which should be the same to the image size
+
+        :return: the sliding window style masks with values ranging between [0, 1].
+        """
+        rows = np.arange(0 + stride - window_size, image_size[0], stride)
+        cols = np.arange(0 + stride - window_size, image_size[1], stride)
+
+        mask_num = len(rows) * len(cols)
+        masks = np.ones((mask_num, image_size[0], image_size[1]),
+                        dtype=np.float64)
```
brianhhu, line 52: dtype can probably be boolean or uint8 if we care about saving memory/storage since the masks here are binary.

xaitk_saliency/utils/masking.py
```diff
@@ -28,3 +29,36 @@ def generate_masked_images(
         masked_img = np.multiply(mask, img, casting='unsafe')
         masked_imgs.append(PIL.Image.fromarray(np.uint8(masked_img)))
     return masked_imgs
+
+
+def weight_regions_by_scalar(
```
brianhhu, line 34: Similarly, there may be a vectorized form of this function to account for multiple classes as the same time.

xaitk_saliency/utils/masking.py
```diff
@@ -28,3 +29,36 @@ def generate_masked_images(
...
+    # Creating an empty heatmap for aggregating weighted matrices
+    heatmap = np.zeros((masks.shape[-2:]))
+
+    # Iterating through perturbation mask and respective weight
+    for weight, prtb_region in zip(scalar_vec, masks):
+        heatmap += ((1 - prtb_region) * weight)
+
+    normalized_heatmap = heatmap / len(scalar_vec)
+    final_heatmap = minmax_scale(normalized_heatmap.ravel(),
+                                 feature_range=(0, 1)).reshape(heatmap.shape)
```
brianhhu, line 63: I'm not sure saliency map normalization is in the scope of the computation or not? We may just want to return "raw" saliency maps to the user, and have them perform normalization as they see fit themselves. I guess that is open for discussion and will also apply to the other saliency implementations.

xaitk_saliency/impls/saliency/fsal.py
brianhhu, line 7: I'm not sure I would call this implementation Fsal, since this is pretty Kitware-specific. It essentially is an occlusion-based saliency algorithm, so maybe something like Occlusion instead?

brianhhu (commented):
In general, I think it looks good for a first pass. One issue I have is that at least from the Jupyter notebook example, it is still very "IQR-focused". For example, we extract features from before the classifier, and feed that into a new logistic regression classifier, which is a common IQR pattern, but not necessarily for classification in general. I think the more common use case of classifier-based saliency would be to try to explain the prediction logits of the model itself (i.e. without needing an additional classifier). So this would require finding the appropriate ImageNet class indices for dog/cat and feeding in those predicted logits to the black box saliency model.

xaitk_saliency/impls/saliency/fsal.py
```diff
@@ -0,0 +1,65 @@
...
+    perturbations (as would be output from a
+    `PerturbImage` implementation.
+    """
+
+    def generate(
+            self,
+            image_conf: np.ndarray,
+            perturbed_conf: np.ndarray,
+            perturbed_masks: np.ndarray
+    ) -> np.ndarray:
```
Purg, line 23: We should not repeat inherited function doc-strings. Pass-through of the doc-string into the leaf implementation is automatically handled. E.g.
```python
import xaitk_saliency

class TestImpl (xaitk_saliency.ImageSimilaritySaliencyMapGenerator):
    def generate(self):
        pass
    def get_config(self):
        pass

t = Test()
help(t.generate)
# outputs parent class doc-string because we didn't override it.
```
This just helps reduce copy-paste errors if something upstream changes and we don't remember to re-copy-paste doc-strings.

xaitk_saliency/impls/saliency/sliding_window.py
```diff
@@ -0,0 +1,120 @@
...
+        mask_shape = [-1] + [*image_size] + [1]
+        masks = masks.reshape(mask_shape)
+        return masks
+
+    def perturb(
+        self,
+        ref_image: PIL.Image.Image
+    ) -> Tuple[List[PIL.Image.Image], np.ndarray]:
+        """
+        Transform an input reference image into a number of perturbed
```
Purg, line 83: See previous comment on not overriding doc-string if there are no changes (shouldn't be for parent class method overrides).

xaitk_saliency/impls/saliency/sliding_window.py
```diff
@@ -0,0 +1,120 @@
...
+        :return: Tuple of perturbed images and the window masks detailing perturbation
+            areas.
+        """
+        img_dims = ref_image.size
+        self.masks = self.generate_block_masks(self.window_size,
+                                               self.stride, img_dims)
+        masked_images = generate_masked_images(self.masks, ref_image)
+        return masked_images, self.masks
+
+    def __call__(
```
Purg, line 113: This is already implemented in the abstract class. Why is this repeated here?

Purg (commented):
We should add unit tests for the functionality added here.

xaitk_saliency/impls/saliency/sliding_window.py
vbhavank: Removed abstract class method left behind from interface copy-paste.

xaitk_saliency/impls/saliency/fsal.py
vbhavank: Removed repeating doc-strings. I did not know this was done automatically. Thank you for the neat example.

xaitk_saliency/impls/saliency/fsal.py
```diff
@@ -0,0 +1,65 @@
+from xaitk_saliency import ImageClassifierSaliencyMapGenerator
```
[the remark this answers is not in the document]
vbhavank: Moved both sliding window and occlusion based generation algorithm impls into folder vis_sal_similarity

pyproject.toml
```diff
@@ -61,8 +61,9 @@ pytest-cov = "^2.11.1"
 ipython = "^7.22.0"
 
 [tool.poetry.plugins."smqtk_plugins"]
+"fsal" = "xaitk_saliency.impls.saliency.fsal"
+"sliding_window" = "xaitk_saliency.impls.saliency.sliding_window"
```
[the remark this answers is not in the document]
vbhavank: Done.

xaitk_saliency/impls/saliency/fsal.py
vbhavank: Agreed, I changed it to OcclusionScoring.

xaitk_saliency/utils/masking.py
vbhavank: That makes sense to me. I moved the scaling into the notebook and modified the weight_regions_by_scalar() to return raw saliency maps.

xaitk_saliency/impls/saliency/sliding_window.py
vbhavank: Done. Changed it to use uint8 masks

xaitk_saliency/impls/saliency/fsal.py
vbhavank: Good point, I vectorized this to perform saliency computation across all classes and regions of the image.

xaitk_saliency/utils/masking.py
vbhavank: Done, I vectorized this to perform a weighted average of all classes and regions of the image.

xaitk_saliency/impls/saliency/fsal.py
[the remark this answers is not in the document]
brianhhu: I'm a little confused by this choice- isn't the sliding window masking technique independent of the final "application area"? For example, the sliding window perturbation can be used for image classification, similarity, detection, etc. Similarly, the current OcclusionScoring method is an ImageClassifierSaliencyMapGenerator, yet it now lives in vis_sal_similarity (which should be using ImageSimilaritySaliencyMapGenerator)?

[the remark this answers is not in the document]
vbhavank: I apologize for the confusion. I should have paid more attention to the submodule names. I added a new module perturber to carry all perturbation-based methods and moved OcclusionScoring to vis_sal_classifier. Please let me know what you think. Thank you!

xaitk_saliency/utils/masking.py
Purg: We defined in the API document that saliency maps return in the [0,1] range. My potential concern with not bounding the output to something is that tools around saliency generator output, when configured with a different implementation, will not be able to handle the variety of unconstrained output.

xaitk_saliency/utils/masking.py
```diff
@@ -28,3 +81,32 @@ def generate_masked_images(
...
         masked_imgs.append(PIL.Image.fromarray(np.uint8(masked_img)))
     return masked_imgs
+
+
+def weight_regions_by_scalar(
+    scalar_vec: np.ndarray,
+    masks: np.ndarray
+) -> np.ndarray:
+    """
+    Weight some binary masks reguib with its respective vector in scalar_vec.
```
nitesh201, line 91: "reguib" is meant to be "region" right? (not sure if this is some specialized XAI terminology or a typo)

vbhavank: Yes, that was a typo. Thank you for catching that!

tests/impls/vis_sal_classifier/test_occlusion_classifier.py
```diff
@@ -0,0 +1,101 @@
...
+class TestOcclusionBasedScoring (TestCase):
+
+    def test_init_(self) -> None:
+        """
+        Test if implementation is usable.
+        """
+        impl = OcclusionScoring()
+        assert impl.is_usable() and isinstance(impl, ImageClassifierSaliencyMapGenerator)
+
+# Below test case does pass but we should have more discussion
```
vbhavank, line 18: This is testing multiple images as input to the function. This addition to the implementation brings in an additional dimension to all input and output args.  @Purg Do you think this is something in line with our plan or something we should discuss more?

pyproject.toml
```diff
@@ -61,6 +61,8 @@ pytest-cov = "^2.11.1"
 ipython = "^7.22.0"
 
 [tool.poetry.plugins."smqtk_plugins"]
+"xaitk_saliency.impls.vis_sal_classifier.occlusion_scoring" = "xaitk_saliency.impls.vis_sal_classifier.occlusion_scoring"
```
nitesh201, line 64: @Purg @vbhavank just a question, but is our preference for the key to be `xaitk_saliency.impls.vis_sal_classifier.occlusion_scoring` or `impls.vis_sal_classifier.occlusion_scoring`

tests/impls/vis_sal_classifier/test_occlusion_classifier.py
Purg: This is currently not represented in the API, so this test would not apply to this current version. I believe we discussed it once but decided to go with the simpler version to start as a balance between imposed complexity between user and implementer. We can revisit this later if usage reveals more advantage to multi-image vectorization.

pyproject.toml
Purg: I assume this is coming from my use of an `"impls...` starting key in my branch. The key needs to be unique just within the package scope (I believe), so since all local impls will be in the same package, I thought it would be shorter to leave off the redundant package name in the config. To that end I think *I* currently prefer just starting with "impl" on the *key* side. The *value* side still needs to be a valid module path, and just *needs* to include the `xaitk_saliency` part.

examples/OcclusionSaliency.ipynb
```diff
@@ -0,0 +1,211 @@
...
+    "from operator import itemgetter\n",
+    "\n",
+    "# Test image to be explained\n",
+    "test_image = 'catdog.png'\n",
+    "# Imagenet class for which saliency maps need to be computed \n",
+    "sal_classes = ['bull mastiff', 'Egyptian cat']\n",
+    "\n",
+    "# Declaring xaitk implementation instantiations \n",
+    "\n",
+    "sliding_window_cfg = {\n",
```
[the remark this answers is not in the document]
Purg: This will also need an update with the class rename in the base branch (you'll need a rebase).

examples/OcclusionSaliency.ipynb
```diff
@@ -0,0 +1,211 @@
...
+      "2021-05-20 16:56:54 (48.0 MB/s) - ‘imagenet_classes.txt.4’ saved [10472/10472]\n",
+      "\n"
+     ]
+    }
+   ],
+   "source": [
+    "\"\"\"\n",
+    "This notebook contains a working example show usage of the new\n",
+    "image classifier based saliency map API for choosing to use \n",
+    "a occlusion based saliency map algorithm.\n",
```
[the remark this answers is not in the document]
Purg: May also want to note that the "warnings" (from other smqtk packages) about psycopg2 and caffe modules missing is expected and OK.

examples/OcclusionSaliency.ipynb
```diff
@@ -0,0 +1,211 @@
...
+    "from torch.autograd import Variable\n",
+    "import torchvision.models as models\n",
+    "from sklearn.linear_model import LogisticRegression\n",
+    "import torch\n",
+    "from sklearn.preprocessing import minmax_scale\n",
+    "import numpy as np\n",
+    "from operator import itemgetter\n",
+    "\n",
+    "# Test image to be explained\n",
+    "test_image = 'catdog.png'\n",
```
Purg, line 64: The other images included here don't seem to be being used. Should they be removed? Or should more cells be added here to show results for all 3 images?

Purg: Separately, what is the license for those images? Are we allowed to bundle them in this repo, or do they need to be downloaded by the user?

tests/impls/vis_sal_classifier/test_occlusion_classifier.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        impl = OcclusionScoring()
+        # Three Pertrubation masks of height and width 10px for 1 class
+        image_confs_1_class_ = np.random.rand(1)
+
+        pertb_confs_1_class_ = np.random.rand(3, 1)
+        mask_confs_1_class_ = np.random.randint(low=0, high=2, size=(3, 10, 10), dtype='int')
+
+        sal = impl.generate(image_confs_1_class_, pertb_confs_1_class_, mask_confs_1_class_)
+        assert sal.shape == (1, 10, 10)
```
Purg, line 49: It might be good to generate expected heatmaps as images or npz files and save those within the `tests/data/` directory for loading and comparison here to ensure results don't regress moving forward.

xaitk_saliency/impls/vis_sal_classifier/occlusion_scoring.py
```diff
@@ -0,0 +1,51 @@
...
+
+        if not (perturbed_masks[0][0][0] in [0, 1]):
+            raise ValueError("Image perturbation mask must be of",
+                             "type integer and binary valued.")
+
+        try:
+            assert len(image_conf) == len(perturbed_conf[0])
+        except AssertionError:
+            raise ValueError("Number of classses in original image and",
+                             " perturbed image do not match.")
```
Purg, line 33: Try-catching an assert is more work than 
```suggestion
        if len(image_conf) != len(perturbed_conf[0]):
            raise ValueError("Number of classses in original image and",
                             " perturbed image do not match.")
```

xaitk_saliency/impls/vis_sal_classifier/occlusion_scoring.py
```diff
@@ -0,0 +1,51 @@
...
+
+        try:
+            assert len(perturbed_conf) == len(perturbed_masks)
+        except AssertionError:
+            raise ValueError("Number of perturbation masks and respective",
+                             "confidence lengths do not match.")
```
Purg, line 39: Try-catching an assert is more work than
```suggestion
        if len(perturbed_conf) != len(perturbed_masks):
            raise ValueError("Number of perturbation masks and respective",
                             "confidence lengths do not match.")
```

xaitk_saliency/impls/vis_sal_classifier/occlusion_scoring.py
Purg: Second observation: We can move these pretty standard checks into the interface and move to a template method pattern w.r.t. the abstract method. This can be a separate PR.

xaitk_saliency/impls/vis_sal_classifier/occlusion_scoring.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def generate(
+            self,
+            image_conf: np.ndarray,
+            perturbed_conf: np.ndarray,
+            perturbed_masks: np.ndarray
+    ) -> np.ndarray:
+
+        if not (perturbed_masks[0][0][0] in [0, 1]):
+            raise ValueError("Image perturbation mask must be of",
```
Purg, line 26: Maybe OK for now, but this may eventually cause issues with other perturbation algorithms that output float-type masks (e.g. RISE). It may be better for this implementation to document a rule in how it handles float-type masks (e.g. rounds valuesl, then binarizes).

xaitk_saliency/interfaces/vis_sal_classifier.py
```diff
@@ -65,8 +65,8 @@ def generate(
             Perturbation masks over the reference image.
             This should be parallel in association to the classification
             results input into the `perturbed_conf` parameter.
-            This should have a shape `[nMasks x H x W]`, be float-typed and
```
Purg, line 68: This should not be just integers (see RISE). We maybe want to "relax" the typing requirement similar to the perturb interface, in that we don't specifically require bool, int or float, but indicate that it *may* be either of those three types and the implementation should handle accordingly.

xaitk_saliency/impls/saliency/sliding_window.py
Purg: This is subsumed by the base branch (which uses bool mask generation).

examples/OcclusionSaliency.ipynb
```diff
@@ -0,0 +1,211 @@
...
+    "slid_algo = from_config_dict(sliding_window_cfg, PerturbImage.get_impls())\n",
+    "occ_alg = from_config_dict(occ_cfg, ImageClassifierSaliencyMapGenerator.get_impls())\n",
+    "\n",
+    "# Creating an example deep feature extractor to describe images\n",
+    "model = models.resnet18(pretrained=True)\n",
+    "model = model.eval()\n",
+    "model = model.cuda()\n",
+    "\n",
+    "# Download ImageNet labels and extract id for required classes\n",
+    "!wget https://raw.githubusercontent.com/pytorch/hub/master/imagenet_classes.txt\n",
```
Purg, line 93: I missed this before, but we should add this file path/name to the `.gitignore` file (or alternatively create one for this directory) to prevent this from getting mistakenly committed to this repo.

xaitk_saliency/utils/masking.py
Purg: In running the example notebook, I can see that the output map is in the state of being not normalized. Did we come to more firm opinions on how this output should take form?

Purg:
@vbhavank Please rebase this branch on the updated master (below hashes according to your currently pushed fork branch)
```bash
git fetch --all
git rebase --onto a733838 203da69 <your local branch name>
```

examples/OcclusionSaliency.ipynb
vbhavank: Deleted the images not being used. 

The image was originally from the MS COCO dataset. Looks like a CCA license. Here are the [terms of use](https://cocodataset.org/#termsofuse). How do we generally handle images in the repo? I can grab a standard Lenna image if that is advised.

examples/OcclusionSaliency.ipynb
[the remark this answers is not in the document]
vbhavank: Add note regarding missing psycopg2 and caffe modules

examples/OcclusionSaliency.ipynb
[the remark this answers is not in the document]
vbhavank: I changed the notebook to construct the implementation directly for simplicity of example.

tests/impls/vis_sal_classifier/test_occlusion_classifier.py
vbhavank: Agreed! I added a condition for checking equality with known values in the next testing function.

xaitk_saliency/impls/vis_sal_classifier/occlusion_scoring.py
vbhavank: Good idea! Sounds good, I will start an issue on making standard checks abstract methods.

xaitk_saliency/impls/vis_sal_classifier/occlusion_scoring.py
vbhavank: Ah, I see. Thank you for those examples. I changed it to the suggested pattern.

xaitk_saliency/interfaces/vis_sal_classifier.py
vbhavank: Modified docstrings indicate that masks can be either one of the three acceptable formats, and implementations are responsible for handling them.

examples/OcclusionSaliency.ipynb
vbhavank: Add a .gitignore for this specific directory.

xaitk_saliency/utils/masking.py
vbhavank: Although normalization is more of a visualization thing and not related to generating process itself, @Purg's point about future conflicts between different versions is valid. The only drawback to always normalizing is a fixed dynamic range of values for better or worse depending upon the implementation, which is not a huge disadvantage from my perspective. @brianhhu let me know what you think. I can revert this change to add the normalization to OcclusionScoring.generate() instead of the notebook.

examples/OcclusionSaliency.ipynb
Purg: The pattern you added is a bit overkill, but it is fine for now.

examples/OcclusionSaliency.ipynb
Purg: It is generally preferred to not have binary content in the repo unless needed for reliability and consistency reasons (e.g. small data files for tests, or reproducible examples). Lenna image is not OK to redistribute (e.g. [see here](https://github.com/matplotlib/matplotlib/issues/4927)). COCO is fine but it's license (CC by 4.0) still requires attribution ([see "Under the following terms" here](https://creativecommons.org/licenses/by/4.0/)).

You're removal here is good enough, however the binary data is still in the git history since you removed it in the latest commit, as opposed to an amended original commit, so the binary files are still present in the object history.

tests/impls/vis_sal_classifier/test_occlusion_classifier.py
```diff
@@ -0,0 +1,83 @@
...
+        impl = OcclusionScoring()
+        assert impl.is_usable() and isinstance(impl, ImageClassifierSaliencyMapGenerator)
+
+    def test_1class_scores(self) -> None:
+        """
+        Test basic scoring with a single class for broadcasting sanity check.
+        """
+        impl = OcclusionScoring()
+        # Three Pertrubation masks of height and width 10px for 1 class
+        image_confs_1_class_ = np.random.rand(1)
```
Purg, line 24: When using random values in unit tests, I'd recommend using seeded generator for reproducibility.

xaitk_saliency/impls/vis_sal_classifier/occlusion_scoring.py
Purg: I already created an issue for the standard checks / template method.

xaitk_saliency/impls/vis_sal_classifier/occlusion_scoring.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+
+    def generate(
+            self,
+            image_conf: np.ndarray,
+            perturbed_conf: np.ndarray,
+            perturbed_masks: np.ndarray
+    ) -> np.ndarray:
+
+        if isinstance(perturbed_masks[0][0][0], float):
```
Purg, line 32: This is a good step. I worry this is overly specific however. Am I correct in understanding that this implementation effectively requires boolean masks? If so we could perform the check `if perturbed_masks.dtype != bool:` to enter the below logic.

xaitk_saliency/impls/vis_sal_classifier/occlusion_scoring.py
```diff
@@ -0,0 +1,61 @@
...
+            # Roundoff when perturbation mask value is of type float
+            perturbed_masks = np.round(perturbed_masks)
+            # Compute binary midpoint for setting threshold
+            binary_midpoint = (np.max(perturbed_masks))/2
+            # Binarizing mask values to be of type int and range [0, 1]
+            perturbed_masks = np.where(perturbed_masks > binary_midpoint,
```
Purg, line 38: I think this line is equivalent to `perturbed_masks = perturbed_masks > binary_midpoint`, which results in a boolean-typed matrix.

vbhavank: You are correct! I changed this to the simpler form suggested.

tests/impls/vis_sal_classifier/test_occlusion_classifier.py
vbhavank: Great point. Added a seed

xaitk_saliency/impls/vis_sal_classifier/occlusion_scoring.py
vbhavank: Agreed! it was not independent of shape and could cause issues for integers that are not 0 and 1. Changed it to your suggestion. Thank you.

brianhhu (commented):
Thanks, it looks better. There is an unused import of sklearn.LogisticRegression in the Jupyter notebook which can probably be removed (carry-over from the more IQR-focused example from before). Also, from the original [Grad-CAM paper](https://arxiv.org/pdf/1610.02391.pdf), the ImageNet classes used are 'boxer' (dog) and 'tiger cat' (cat).

xaitk_saliency/impls/vis_sal_classifier/occlusion_scoring.py
```diff
@@ -0,0 +1,60 @@
...
+
+        if perturbed_masks.dtype != bool:
+            # Roundoff when perturbation mask value is of type float
+            perturbed_masks = np.round(perturbed_masks)
+            # Compute binary midpoint for setting threshold
+            binary_midpoint = (np.max(perturbed_masks))/2
+            # Binarizing mask values to be of type int and range [0, 1]
+            perturbed_masks = perturbed_masks > binary_midpoint
+            warnings.warn("Image perturbation mask must be of type integer,"
+                          "masks are rounded and binarized.", UserWarning)
```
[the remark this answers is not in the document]
Purg: Mechanically speaking, performing the conversion does not preclude the input of differently typed masks (literally the point of the conversion code being here). If there was a boolean requirement *and NO conversion*, that would be a situation where this impl would not be able to handle RISE-generated inputs. I believe what I see here would "work" for both boolean, int and float typed inputs.

I think there is a separate question about whether this generator algorithm *logically* needs boolean-ified inputs or if it should make use of more-than-bool masks, but I think that's a question separate (larger) from this region of conversion code. We can also choose to address and answer that question in a separate issue/pull-request. I don't want to spin our wheels overly long on these initial implementation integrations beyond an initial level of "they work." Making them "work better" is a separate task.

tests/impls/vis_sal_classifier/test_occlusion_classifier.py
Purg: Looks good. I don't request the following, but FYI there is another method of stabilly performing seeded random operations isloated from other uses of random: np.random.RandomState(seed: int) which results in an object that you can do all the normal random function calls on. This can isolate the random seeding to that new state and not affect other, e.g. nested, uses of random.

xaitk_saliency/impls/vis_sal_classifier/occlusion_scoring.py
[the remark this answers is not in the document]
brianhhu: Sounds good, I don't want to hold up this PR much longer. I was mainly thinking from the longer-term perspective of compatibility with non-binary masks, e.g. [RISE](https://github.com/eclique/RISE/blob/master/explanations.py#L56) where the saliency computation might be changed (or possibly even incorrect) if you convert the masks to be binary.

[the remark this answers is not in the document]
Purg: Valid. We currently state in the abstract interface that "a value closer to 1.0 indicate areas of the image that are *unperturbed*" which maybe assumes the interpretation that values in between 0 and 1 are to be considered (linearly?) partial perturbations. Of course, what is a "partial perturbation" or even a "perturbation" in the strict sense, which we could define or could turn out to be a quagmire of over-specification.

At the very least, the mask generation implementation should document it's assumptions about input masks, for where they are aware assumptions are being made. In this case, for example, that masks are (currently) converted to binary based on a computed midpoint value.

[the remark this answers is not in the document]
vbhavank: Based on the offline conversations, I have removed the type check for masks.

xaitk_saliency/utils/masking.py
```diff
@@ -83,3 +84,35 @@ def generate_masked_images(
...
+    :return: A numpy array representing the weighted heatmap.
+    """
+    # Weighting each perturbed region with its respective score in vector.
+    heatmap = (np.expand_dims(np.transpose(1 - masks), axis=3) * scalar_vec)
+
+    # Aggregate scores across all perturbed regions.
+    sal_across_masks = np.transpose(heatmap.sum(axis=2))
+
+    # Compute final saliency map by normalizing with sampling factor.
+    final_heatmap = sal_across_masks/masks.sum(axis=0)
```
vbhavank, line 116: @nitesh201 Please take a look at the newer version accounting for any unequal sampling by normalizing the final saliency map with the aggregate of all perturbed masks. Feel free to give this a try with RISE, and please let me know if you come across issues.

xaitk_saliency/utils/masking.py
```diff
@@ -83,3 +84,32 @@ def generate_masked_images(
...
+                       [nMasks, nClasses]
+    :param masks: Mask array in the [nMasks, Height, Weight] shape format.
+
+    :return: A numpy array representing the weighted heatmap.
+    """
+
+    heatmap = ((1 - np.transpose(masks))[..., newaxis] * scalar_vec)
+
+    # Computing average scores across all perturbations for nClasses
+    final_heatmap = heatmap.sum(axis=2) / len(scalar_vec)
```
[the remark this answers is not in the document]
vbhavank: Changed this to normalize according to perturbed masks.

tests/impls/vis_sal_classifier/test_occlusion_classifier.py
```diff
@@ -0,0 +1,84 @@
...
+    def test_standard_1class_scores(self) -> None:
+        """
+        Test basic scoring on known values and non-square masks.
+        """
+        impl = OcclusionScoring()
+        # Three Pertrubation masks of size 4 x 6 for 1 class
+        image_confs_1_class_ = np.array([0.6])
+        pertb_confs_1_class_ = np.array([[0.3], [0.65], [0.12], [0.18], [0.36], [0.42]])
+        sal = impl.generate(image_confs_1_class_, pertb_confs_1_class_, EXPECTED_MASKS_4x6)
+        standard_sal = np.load('tests/data/OccScorSal.npy')
```
Purg, line 41: FYI, the test data directory path is defined in `tests/__init__.py` as an importable attribute. This is OK for now, but is a little fragile to where you run your test, which is currently not being upheld in CI. This line will probably be updated in the future.

Purg (approved):
I think this looks good to do from my perspective.

vbhavank:
> Thanks, it looks better. There is an unused import of sklearn.LogisticRegression in the Jupyter notebook which can probably be removed (carry-over from the more IQR-focused example from before). Also, from the original [Grad-CAM paper](https://arxiv.org/pdf/1610.02391.pdf), the ImageNet classes used are 'boxer' (dog) and 'tiger cat' (cat).

Thank you. Corrected classes to 'boxer' and 'tiger car'.

Purg (changes requested):
Oh! I almost forgot: please add some change notes! Please add one or more change notes to the `docs/release_notes/pending_release.rst` file.

vbhavank:
> Oh! I almost forgot: please add some change notes! Please add one or more change notes to the `docs/release_notes/pending_release.rst` file.

Added a single-line description to pending release notes. Please let me know if you prefer any additional details.

docs/release_notes/pending_release.rst
```diff
@@ -5,9 +5,9 @@ XAITK-Saliency Pending Release Notes
 Updates / New Features
 ----------------------
 
-Interfaces
+Implementations
```
Purg, line 8: You seem to have *replaced* the contents of this file. Your notes should be additions, instead.

Purg (approved):

docs/release_notes/pending_release.rst
vbhavank: Sorry I was not sure how to fill that up. I reverted my commit and made it an addition. Please let me know what you think.

xaitk_saliency/utils/masking.py
brianhhu: Per our offline discussion, line 115 should be

```
final_heatmap = sal_across_masks/(1 - masks).sum(axis=0)
```

Otherwise, everything else LGTM!

brianhhu (approved):

PR #7, opened 2021-05-18
Replace logit-sal "augmentor" with OcclusionBasedPerturb impl

Purg:


xaitk_saliency/utils/masking.py
```diff
@@ -1,12 +1,64 @@
-from typing import List
+from typing import List, Tuple
 
 import numpy as np
 import PIL.Image
 
 
+def generate_block_masks(
```
Purg, line 7: @brianhhu @vbhavank There is a different implementation of this in the smqtk mono-repo dev/XAI branch [here](https://github.com/Kitware/SMQTK/blob/dev/XAI/python/smqtk/utils/saliency.py#L112). Is this implementation a valid or more optimal strict replacement for this function

xaitk_saliency/impls/perturb_image/occlusion.py
```diff
@@ -0,0 +1,64 @@
...
+    def __init__(self, window_size: int = 50, stride: int = 20):
+        self.window_size = int(window_size)
+        self.stride = int(stride)
+
+    def perturb(
+        self,
+        ref_image: PIL.Image.Image
+    ) -> Tuple[List[PIL.Image.Image], np.ndarray]:
+        ref_mat = np.asarray(ref_image)
+        masks = mask_lhs = generate_block_masks(
```
Purg, line 44: I realize that this is a change from the base implementation that moves the mask generation into the runtime method, however the previous implementation had a requirement that all input imagery match the generated mask size. This is a bit fragile as data in the wild is often of different sizes.

We can do something where we use result caching (`functools.lrucache`) to optimize this for repeat image size invocations.

Alternatively, is there instead a desire for this impl to generate a static mask grid and then "scale" that to the input imagery? This impl will create blocks of constant N-pixel squares. The percent impact on input imagery will however depend on the scale of the input. E.g. an image scaled to NxM will be masked differently than if it was scaled to (5*N)x(5*M).

xaitk_saliency/utils/masking.py
brianhhu: I'm not sure about this- it looks like this code was mainly Alina's work? I haven't worked with this utils code before.

Purg: The linked file *and* this function in `Logit_ImageSaliencyAugmenter` are both attributed to Alina via git-blame

xaitk_saliency/impls/perturb_image/occlusion.py
brianhhu: My understanding is that mask generation is somewhat of a bottleneck (i.e. we currently generate ~2K masks using nested for loops). I don't know what kind of performance hit we should expect if we have to regenerate the masks for each input image (although caching would probably help). Another hypothetical scenario would be for extremely high-resolution imagery- it might not be feasible to fit the full-resolution masks in memory. For our use case (which I think is pretty standard), input images were generally assumed to be re-sized to a standard size (e.g. 224x224 pixels).

I didn't quite understand the resizing argument. Either way, aren't we either resizing the masks or resizing the image, and I imagine there is potentially some equivalence there? In the case of re-sizing the mask, would we use some type of nearest neighbors interpolation? Other linear interpolation schemes would make the masks non-binary.

Purg: I assume we don't want to resize the image since the current API states the output perturbations are the same resolution as the input. The argument there was so that we don't have implementations outputting potentially different image perturbations such that their scales are different than inputs for next-step consistency.

Let me try to add a more reasonable scale unit test for performance evaluation.

xaitk_saliency/utils/masking.py
vbhavank: I believe the difference between the two implementations is the consideration of image pixels at image boundaries. From what I remember, the linked file's function ignores boundary pixels if it cannot fit a complete sliding window, while Logit_ImageSaliencyAugmenter is the right one that factors in boundary pixels. On a related note, this seems to be a stray function that was refactored [here](https://github.com/Kitware/SMQTK/blob/dev/XAI/python/smqtk/algorithms/saliency/sbsm.py#L173) with additional sanity checks.

Purg:
Last evening I experimented with a streaming version of the perturb function. I found that this streaming variant is about 1.7-2x faster than the original block-allocation method. I attempted a number of different internal variants as well until I settled on the current form I have. I have not tried utilizing streamed parallelism.
```python
In [1]: from xaitk_saliency.impls.perturb_image.occlusion import OcclusionBasedPerturb
   ...: import numpy as np
   ...: import PIL.Image
   ...: inst = OcclusionBasedPerturb(24, 5)
   ...: white_image = PIL.Image.fromarray(np.ones((256, 512, 3), dtype=np.uint8) * 255)

In [2]: %timeit inst.perturb(white_image)  # REQUIRES at block memory allocation, 2+GB this test case
6.04 s ± 31.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [3]: %timeit list(inst.perturb_stream(white_image))  # variable mem requirement depending on how app wants to handle results
3.51 s ± 9.96 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [4]: %%timeit
   ...: # Version that does not block store results, Note that we're faster still due to not allocating all that memory.
   ...: shapes = []
   ...: for img, m in inst.perturb_stream(white_image):
   ...:     shapes.append( (img.size, m.shape) )
   ...: 
   ...: 
3.02 s ± 66 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

No, this variant is not compatible with the current interface API. If we decided to go this way we need to update the interface API (or at the very least add to it).

Purg:
I pulled in my streaming parallel map utility and have the following results with threading (multiprocessing is strictly worse). 4 threads were configured, with no additional returns using more than that in this test case
```python
In [1]: from xaitk_saliency.impls.perturb_image.occlusion import OcclusionBasedPerturb
   ...: import numpy as np
   ...: import PIL.Image
   ...: inst = OcclusionBasedPerturb(24, 5)
   ...: white_image = PIL.Image.fromarray(np.ones((256, 512, 3), dtype=np.uint8) * 255)

In [2]: %%timeit
   ...: shapes = []
   ...: for img, m in inst.perturb_stream2(white_image):
   ...:     shapes.append( (img.size, m.shape) )
   ...: 
   ...: 
960 ms ± 12.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

xaitk_saliency/utils/masking.py
Purg: I'm going to leave this as is for now baring specific need to update. That will probably come with the shift to a streaming API that we've talked about.

Purg:
Rolled the experimental commit back for the purposes of getting this PR approved for merger. Streaming consideration will come in a subsequent PR.

xaitk_saliency/impls/perturb_image/occlusion.py
Purg: @brianhhu Wanted to circle back on this after our discussion about investigating a streaming API going forward. Is this OK for this initial PR?

xaitk_saliency/impls/perturb_image/occlusion.py
```diff
@@ -0,0 +1,64 @@
...
+        ...      [[255, 255, 255], [255, 255, 255]]]
+        ... )
+
+    :param window_size: The block window size.
+    :param stride: The sliding window striding step.
+    """
+
+    def __init__(self, window_size: int = 50, stride: int = 20):
+        self.window_size = int(window_size)
+        self.stride = int(stride)
```
brianhhu, line 37: Do we plan to only support square window sizes and symmetric strides? I imagine this covers 95% of use cases, but I can potentially image where non-square occluders (e.g. rectangles) or non-symmetric strides (i.e. different in x,y) might be useful. I see other libraries either allow ints or tuples of ints to account for this.

xaitk_saliency/impls/perturb_image/occlusion.py
brianhhu: Yes, I think this is fine for now- thanks.

brianhhu (commented):
In general, it looks good to me- one question I had was in regards to how we plan to deal with the concurrent "old" and "new" interfaces. Currently, both seem to exist in the package, but do we plan to go with just the "new" interfaces (i.e. vis_sal_*) once we publish this?

xaitk_saliency/impls/perturb_image/occlusion.py
brianhhu: I think this is fine for now- thanks.

xaitk_saliency/impls/perturb_image/occlusion.py
```diff
@@ -0,0 +1,64 @@
+from typing import Dict, Any, Iterable, Tuple, List
+
+import PIL.Image
+import numpy as np
+
+from xaitk_saliency.interfaces.perturb_image import PerturbImage
+from xaitk_saliency.utils.masking import generate_block_masks
+
+
+class OcclusionBasedPerturb (PerturbImage):
```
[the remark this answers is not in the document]
Purg: Sure! That name sounds fine/more appropriate here.

xaitk_saliency/impls/perturb_image/occlusion.py
Purg: I don't want to get too far into the weeds of possible use-cases, but this seems sensible and simple enough. I'll take a stab at adding that here, circling back if the change is more of a sink than expected.

vbhavank: Great point, I have come across use cases with non - square occluders generally used as synthetic distractors to strengthen training.

xaitk_saliency/utils/masking.py
```diff
@@ -1,12 +1,64 @@
...
+
+    :param window_size: the block window size (with value 0, other areas
+        with value 1)
+    :param stride: the sliding step
+    :param image_size: the mask size which should be the same to the image
+        size
+
+    :return: the sliding window style masks
+    """
+    rows = np.arange(0 + stride - window_size, image_size[0], stride)
```
nitesh201, line 30: Not 100% aware of the logic of the sliding window occlusions, but wanted to confirm that the following (apparent) behaviors were expected:
a) if stride > window_size, then the points along the perimeter of the masks [where i, j < (stride - window_size)] will not be 0 in any of the masks
b) if stride < window_size, then the set of points where r and c are both negative will all produce identical masks [ranging from (0, 0) to (window_size, window_size)]

I might be wrong, but I can double check this a bit more closely in case it's of interest

nitesh201 (commented):
Looks good to me, had one question about possible edge cases (literally :smile:) in the `generate_block_masks` util function

xaitk_saliency/utils/masking.py
nitesh201: If I understand the map generation process correctly, I'm curious if undersampling/oversampling the boundaries could potentially lead to saliency clustering or dispersing towards the edges.

Purg: I'm not sure I'm fully parsing your question. However, I would say that this is a separate issue that we should address in a separate branch/PR after getting this initial implementation in.

vbhavank (approved):
Looks good to me!

xaitk_saliency/utils/masking.py
nitesh201: basically I'm just asking about how the ranges set in lines 30/31 affect the masks set in line 54

nitesh201 (approved):

xaitk_saliency/utils/masking.py
vbhavank: @nitesh201 I believe you are right. In condition, a) boundary pixels might not be masked and I expect this would cause the saliency values to be focused around the centre of the image.

nitesh201: Main point being that the statement "Generates a number of mask images with each masking a distinct square." seems to be violated under certain conditions

Purg: :+1: I'm going to "resolve" this for now since this is appropriate for a follow-on branch/PR. This one is more simply introducing the structure and carrying the previous Phase2 implementation forward. Thank you for catching this though!

Purg:
> In general, it looks good to me- one question I had was in regards to how we plan to deal with the concurrent "old" and "new" interfaces. Currently, both seem to exist in the package, but do we plan to go with just the "new" interfaces (i.e. vis_sal_*) once we publish this?

@brianhhu To your question, I did remove the "old" implementation, `LogitImageSaliencyAugmenter`, that this new implementation supersedes. I didn't remove the "old" *interface* yet since the RISE algorithm is still to be transitioned in a separate branch. The plan is to go with just the "new" interfaces and implementations.

xaitk_saliency/impls/perturb_image/occlusion.py
Purg: The new update should have tuple inputs for window and stride and an added unit-test that tests a rectangular window/stride example.

brianhhu: Looks good to me- thanks for making the change.

brianhhu:
Sounds good, this looks good to me- feel free to merge when ready.

PR #8, opened 2021-05-18
Update SMQTK-* deps to the versions that do not require setuptools

Purg:
Also update other deps via a `poetry update`.


Purg:
@vbhavank This should address the issue you had last week when utilizing a conda environment with a different/older setuptools package.

PR #9, opened 2021-05-25
Transition RISE perturbation to new api

nitesh201:
A previous implementation of the RISE perturbation algorithm described [here](https://arxiv.org/pdf/1806.07421.pdf) is transitioned to conform to the XAITK saliency API.


nitesh201:
@Purg The new update, which upsamples masks on `perturb()` in order to handle images of multiple sizes with the same RISEPerturbation, has two implications that may require slight refactoring of the notebook branch:

1) there is no longer an `input_size` parameter to the `__init__` function
2) due to random cropping in the upsampling operation, the results of `perturb()` are no longer idempotent

xaitk_saliency/impls/perturb_image/rise.py
```diff
@@ -0,0 +1,90 @@
...
+
+        self.grid = grid
+
+    def perturb(
+        self,
+        ref_image: PIL.Image.Image,
+    ) -> Tuple[List[PIL.Image.Image], np.ndarray]:
+        ref_mat = np.asarray(ref_image)
+        input_size = (ref_image.height, ref_image.width)
+        masks = self._generate_masks(input_size=input_size)
```
brianhhu, line 70: For RISE, we have an internal function which generates masks, while for the sliding window approach, we have a utility function in xaitk_saliency.utils.masking. Is one approach preferred over the other?

xaitk_saliency/impls/perturb_image/rise.py
```diff
@@ -0,0 +1,90 @@
...
+    def __init__(
+        self,
+        N: int,
+        s: int,
+        p1: float,
+        seed: Optional[float] = None
+    ):
+        """
+        Generate a set of random binary masks
+        :param int N:
```
Purg, line 41: Don't need type info in `:param` annotations since they're in the actual function annotations.

brianhhu (commented):
In general, LGTM- should probably have some additional pointers to the RISE paper and associated repository.

xaitk_saliency/impls/perturb_image/rise.py
```diff
@@ -0,0 +1,90 @@
...
+            Number of random masks used in the algorithm. E.g. 1000.
+        :param int s:
+            Spatial resolution of the small masking grid. E.g. 8.
```
Purg, line 44: Might be good to note that this is a "side length" and the resulting grid is square.

xaitk_saliency/impls/perturb_image/rise.py
Purg: I won't say no to a utility function.
Usually if a function is usable in multiple locations, then a utility function is called for.
However if the function is implementation specific and only used in one location, I think it's find if it only exists in that location.

Purg: I will also say that since this function is not being utilized anywhere else other than this line, it might as well be in-line. I won't require that change specifically.

xaitk_saliency/impls/perturb_image/rise.py
```diff
@@ -0,0 +1,90 @@
...
+        if ref_mat.ndim > 2:
+            mask_lhs = masks.reshape(-1, *input_size, 1)
+        masked_images = (mask_lhs * ref_mat).astype(ref_mat.dtype)
+        masked_images_pil = [
+            PIL.Image.fromarray(
+                m,
+                mode=ref_image.mode
+            )
+            for m in masked_images
+        ]
```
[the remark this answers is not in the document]
Purg: I won't complain about consistency, but what aspect were you concerned was inconsistent?

When moving to a streaming API this block will be refactored anyway.

tests/impls/perturb_image/test_rise.py
```diff
@@ -0,0 +1,142 @@
...
+            assert img.mode == "RGB"
+            # Test for expected output perturbed image content
+            expected_img_ch = EXPECTED_IMAGES_4x6_L[i]
+            expected_img = np.stack(
+                (expected_img_ch, expected_img_ch, expected_img_ch),
+                -1
+            )
+            assert np.allclose(img, expected_img)
+
+    def test_perturb_4channel(self) -> None:
```
Purg, line 93: If this has no content this might as well be removed.

tests/impls/perturb_image/test_rise.py
```diff
@@ -0,0 +1,142 @@
...
+        images of varying sizes
+        """
+        impl = RISEPertubation(N=2, s=2, p1=0.5, seed=42)
+        white_image_small = PIL.Image.fromarray(
+            np.full((4, 6), fill_value=255, dtype=np.uint8)
+        )
+        white_image_large = PIL.Image.fromarray(
+            np.full((41, 26), fill_value=255, dtype=np.uint8)
+        )
+        _, _ = impl.perturb(white_image_small)
```
Purg, line 108: Might want to add asserts that the outputs are of the expected shapes based on the input.

xaitk_saliency/impls/perturb_image/rise.py
```diff
@@ -0,0 +1,90 @@
...
+            this resolution to be applied to (multiplied with) the input later.
+            E.g. (224, 224).
+        """
+        masks = np.empty((self.N, *input_size))
+        cell_size = np.ceil(np.array(input_size) / self.s)
+        # Upscale factor
+        up_size = (self.s + 1) * cell_size
+        for i in range(self.N):
+            # Random shifts
+            x = self.rng.integers(0, cell_size[0])
```
Purg, line 24: >     2. due to random cropping in the upsampling operation, the results of `perturb()` are no longer idempotent

You could initialize a random generator every time within this function when there is a provided seed (or even a separate seed from that used in grid generation) to provide optionally deterministic behavior.

Purg: This way you also don't have to retain the `self.rng` instance.

xaitk_saliency/impls/perturb_image/rise.py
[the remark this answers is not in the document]
brianhhu: In general the code is the same, but there is some typing/formatting that is different (really a minor issue). Agreed that this will probably be refactored after introducing the streaming API.

xaitk_saliency/impls/perturb_image/rise.py
brianhhu: Sounds good- I just wanted to get some clarification about software best practices for when and when not to use utility functions. Thanks!

xaitk_saliency/impls/perturb_image/rise.py
```diff
@@ -0,0 +1,90 @@
...
+    ):
+        """
+        Generate a set of random binary masks
+        :param int N:
+            Number of random masks used in the algorithm. E.g. 1000.
+        :param int s:
+            Spatial resolution of the small masking grid. E.g. 8.
+        :param float p1:
+            Probability of the grid cell being set to 1 (otherwise 0). E.g. 0.5.
+            Assumes square grid.
```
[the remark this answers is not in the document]
nitesh201: yeah, this seems to be meant for the s param (i.e. that with s=8 we generate a 8x8 grid)

xaitk_saliency/impls/perturb_image/rise.py
```diff
@@ -0,0 +1,90 @@
...
+                                    anti_aliasing=False)[x:x + input_size[0], y:y + input_size[1]]
+        # Reshape brings this to: (N, input_size[0], input_size[1], 1)
+        return masks
+
+    def __init__(
+        self,
+        N: int,
+        s: int,
+        p1: float,
+        seed: Optional[float] = None
```
Purg, line 37: I just noticed this should be an integer. `np.random.default_rng` expects either an integer or sequence of integers.

tests/impls/perturb_image/test_rise.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        impl = RISEPertubation(N=2, s=2, p1=0.5, seed=42)
+        white_image_small = PIL.Image.fromarray(
+            np.full((4, 6), fill_value=255, dtype=np.uint8)
+        )
+        white_image_large = PIL.Image.fromarray(
+            np.full((41, 26), fill_value=255, dtype=np.uint8)
+        )
+        pert_imgs_small, pert_masks_small = impl.perturb(white_image_small)
+        assert pert_imgs_small[0].size == white_image_small.size
```
Purg, line 106: A simple assert missing here is that of checking the quantity of masks being returned:
```suggestion
        pert_imgs_small, pert_masks_small = impl.perturb(white_image_small)
        assert len(pert_imgs_small) == len(pert_masks_small) == 2  # `N` value above.
        assert pert_imgs_small[0].size == white_image_small.size
```

xaitk_saliency/impls/perturb_image/rise.py
Purg: This will be changed with the streaming update, not going to require any changes here regarding this.

Purg (approved):

brianhhu (approved):
LGTM!

vbhavank (approved):

PR #10, opened 2021-05-27
Update RTD build requirements file

Purg:
I am also trying to get the RTD-build-on-PRs thing working with this PR.


brianhhu (commented):
LGTM, I don't know much about automatic doc generation, so I'll defer to you and others.

Purg:
> LGTM, I don't know much about automatic doc generation, so I'll defer to you and others.

This was more a fix required for automatic building by RTD, otherwise I would have "activated" RTD hosting without needing an update. I should have noted that I was mostly looking for a sanity check / habitual review, though I suppose there wasn't much of that here to be done. :) Thanks for taking a quick look.

PR #11, opened 2021-06-01
Dev/update notebook example

Purg:
Update the occlusion saliency example notebook for flow and demo-ability.

This also fixes an issue with the sliding window perturb-er plugin, whose module path was *not* changed when the file was renamed.

NOTE: This is currently a franken-branch that is based on (read: requires) ~~#6~~ and #9.


examples/OcclusionSaliency.ipynb
```diff
@@ -155,23 +287,28 @@
...
+    "pm_sum = pert_masks.sum(axis=0)\n",
+    "\n",
+    "for i, (class_sal_map, label) in enumerate(zip(class_sal_maps, sal_class_labels)):\n",
     "    plt.subplot(1, sub_plot_ind, i+2)\n",
-    "    plt.imshow(test_img, alpha=0.5)\n",
-    "    plt.imshow(class_sal_map, cmap='jet', alpha=0.3)\n",
-    "    plt.title('{}'.format(sal_classes[i]))\n",
+    "    plt.imshow(test_image, alpha=0.7)\n",
+    "    plt.imshow(\n",
+    "        class_sal_map / pm_sum,\n",
```
Purg, line 308: @vbhavank Pinging you to review this part of the notebook update by request from Brian. I believe that I added this via your direction. Could you confirm that this is appropriate here + why? Brian expressed that the denominator cancels out from the `plt.imshow` auto-normalization.

examples/OcclusionSaliency.ipynb
```diff
@@ -1,12 +1,9 @@
...
  "cells": [
   {
-   "cell_type": "code",
-   "execution_count": 1,
+   "cell_type": "markdown",
    "metadata": {},
-   "outputs": [],
    "source": [
-    "\"\"\"\n",
     "This notebook contains a working example show usage of the new\n",
```
brianhhu, line 7: "This notebook contains a working example **to** show..."

examples/OcclusionSaliency.ipynb
```diff
@@ -19,81 +16,101 @@
     "- torch\n",
```
brianhhu, line 16: Note "mathplotlib" should be "matplotlib" in the package list.

examples/OcclusionSaliency.ipynb
```diff
@@ -19,81 +16,101 @@
...
     "\n",
-    "# Creating an example deep feature extractor to describe images\n",
     "model = models.resnet18(pretrained=True)\n",
     "model = model.eval()\n",
-    "model = model.cuda()\n",
+    "if CUDA_AVAILABLE:\n",
+    "    model = model.cuda()\n",
+    "\n",
+    "# These are some simple helper functions to perform prediction with this model\n",
+    "model_input_size = 224\n",
```
brianhhu, line 69: Probably better to be explicit here:

```
model_input_size = (224, 224)
```

examples/OcclusionSaliency.ipynb
```diff
@@ -19,81 +16,101 @@
...
+    "    transforms.ToTensor(),\n",
+    "    transforms.Normalize(\n",
+    "        mean=[0.485, 0.456, 0.406],\n",
+    "        std=[0.229, 0.224, 0.225]\n",
+    "    ),\n",
+    "])\n",
+    "\n",
+    "\n",
+    "def image_loader(image):\n",
+    "    image = model_loader(image).float()\n",
```
brianhhu, line 81: Minor point: Is the float conversion necessary here?

examples/OcclusionSaliency.ipynb
```diff
@@ -19,81 +16,101 @@
...
+    "    if CUDA_AVAILABLE:\n",
+    "        image = image.cuda()\n",
+    "    return image.unsqueeze(0)\n",
+    "\n",
+    "\n",
+    "def predict_with_model(model, image_tensor):\n",
+    "    featureVec = model(image_tensor)\n",
+    "    # Converting feature extractor output to probabilities.\n",
+    "    return torch.nn.functional.softmax(featureVec, dim=1).cpu().detach().numpy().squeeze()\n",
```
brianhhu, line 90: Minor point: Is the detach necessary here? Since we run the model in eval mode, I don't think gradients are computed.

examples/OcclusionSaliency.ipynb
```diff
@@ -102,38 +119,56 @@
...
-    "    featureVec = model(image_tensor)\n",
-    "    # Converting feature extractor output to probabilities.\n",
-    "    preds_ = torch.nn.functional.softmax(featureVec, dim=1).cpu().detach().numpy().squeeze()\n",
-    "    # Returning scores for only desired class IDs\n",
-    "    return np.asarray(itemgetter(*required_class_id)(preds_))\n",
+    "perturber = SlidingWindowPerturb(\n",
+    "    window_size=(24, 24),\n",
+    "    stride=(4, 4),   # evenly divides the window size per-axis.\n",
+    ")\n",
+    "# perturber = RISEPertubation(\n",
```
brianhhu, line 134: I'm wondering if it's better practice to actually run both algorithms (sliding window and RISE) and compare them side-by-side. One of the benefits of notebooks is that we can visualize the output results of both.

examples/OcclusionSaliency.ipynb
```diff
@@ -143,9 +178,106 @@
...
+    "# Generate the additional imagery for blackbox prediction, and get predictions on those.\n",
+    "pert_images, pert_masks = perturber(test_image)\n",
+    "len(pert_images), pert_masks.shape"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {},
+   "source": [
+    "# This visualization shows the differences in how much regions are masked vs. not masked.\n",
```
brianhhu, line 200: This line renders weirdly in the Jupyter notebook, it may have been meant as a comment.

examples/OcclusionSaliency.ipynb
Purg: Apparently it is. I tried removing it but for some reason I have yet to understand the output tensor is marked for backprop, despite tests ensuring the input is a variable not for grad and that the network is indeed in eval mode.

examples/OcclusionSaliency.ipynb
Purg: This is supposed to be a "raw" cell, e.g. it just looks like a raw text block in jupyter notebook. Google colab and github fail to handle raw cells appropriately.

brianhhu (commented):
LGTM overall, I know we'll probably end up making edits as the RISE/streaming API gets nailed down. One high-level suggestion would be to potentially show both sliding window and RISE results explicitly (instead of asking the user to comment that out).

examples/OcclusionSaliency.ipynb
brianhhu: Weird, I can't argue with that then...

examples/OcclusionSaliency.ipynb
Purg: The original intention was to show an "application" like work flow where one is the chosen path. In the demo that used this notebook, we shows that we could change the part of that once cell that sets the perturber, by direct assignment or by configuration, to something different without impacting the rest of the "application."

examples/OcclusionSaliency.ipynb
vbhavank: I do believe we should be normalizing with pm_sum. The normalization that is being achieved in the plt.imshow was included [here](https://github.com/Purg/xaitk-saliency/blob/dev/update-notebook-example/xaitk_saliency/utils/masking.py#L115). Therefore, re-normalizing would cancel out.

Purg: I read two different conclusions there: "we should be normalizing with pm_sum" and "re-normalizing would cancel out".

vbhavank: @Purg, I apologize. I meant to say I do *not* believe we should be normalizing pm_sum.

Purg: OK, sounds good. I will remove this denominator.

examples/OcclusionSaliency.ipynb
Purg: I am removing this and a few other packages here because they are already depedencies of `xaitk-saliency` and should already be installed if they are running this notebook.

examples/OcclusionSaliency.ipynb
Purg: Nope, functionally not necessary by empirically inspecting just now.

examples/OcclusionSaliency.ipynb
Purg: I have pushed an additional commit (for ease of rollback purposes) that attempts a revision of the framing and narrative of the notebook. Please take a look and see if this better shows what we want here. When we settle I will squash the experiments here to a single commit as the unified "change."

examples/OcclusionSaliency.ipynb
```diff
@@ -1,151 +1,301 @@
...
+    "    print(f\"Saliency maps: {sal_maps.shape}\")\n",
+    "    \n",
+    "    # Visualize mask image region coverage\n",
+    "    mask_coverage = np.zeros(pert_masks[0].shape)\n",
+    "    for pm in pert_masks:\n",
+    "        mask_coverage += pm\n",
+    "    mask_coverage /= len(pert_masks)\n",
+    "\n",
+    "    plt.figure(figsize=(4, 4))\n",
+    "    plt.imshow(mask_coverage)\n",
```
brianhhu, line 134: Maybe don't use autoscaling here when showing mask coverage? It amplifies very small differences when it should be relatively uniform for sliding window and RISE.

Also- I think for RISE specifically, we need to divide mask_coverage by the expectation (i.e. p=0.5). That will make the mask coverage values closer to 1, which corresponds to the sliding window approach.

examples/OcclusionSaliency.ipynb
```diff
@@ -1,151 +1,301 @@
...
+    "This example will follow an application-like use-case where we define a functionally rigid process that transforms an input image into a number of saliency heat-maps based on our black-box classifier's output, visualizing them over the input image.\n",
+    "We will show that it is easy to change which of our API implementations are used in the application without impacting the application's successful execution, first using a sliding-window perterbation algorithm and then using the RISE perturbation algorithm.\n",
+    "\n",
+    "This will necessarilly include the use of a classification model to perform the role of the black-box classifier that we will be determining the saliency for.\n",
+    "We will fill this role here with a PyTorch Imagenet-pretrained ResNet18 network.\n",
     "\n",
-    "Packages required for the below example\n",
+    "### Table of Contents\n",
+    "* [The test image](#The-test-image)\n",
+    "* [The \"application](#The-\"application\")\n",
```
brianhhu, line 18: Missing end quotes:

```
[The \"application\"]
```
The links in the TOC don't render for me in Github, but maybe they do in an actual notebook.

examples/OcclusionSaliency.ipynb
brianhhu: Looks good to me- I added a minor comment about maybe not using auto-normalization when showing mask coverage, since it amplifies small differences and puts sliding window and RISE on two different axes. Thanks!

examples/OcclusionSaliency.ipynb
Purg: I want to say that is a Github issue as this is clearly rendering and working in my jupyter-notebook instance. Below are my installed jupyter versions if anything looks off / non-standard:
```bash
$ jupyter --version
jupyter core     : 4.7.1
jupyter-notebook : 6.4.0
qtconsole        : 5.1.0
ipython          : 7.23.1
ipykernel        : 5.5.5
jupyter client   : 6.1.12
jupyter lab      : not installed
nbconvert        : 6.0.7
ipywidgets       : 7.6.3
nbformat         : 5.1.3
traitlets        : 5.0.5
```

examples/OcclusionSaliency.ipynb
Purg: The application is intentionally not implementation specific. Maybe a "solution" is we not have that mask region coverage part in there at all as it seems to be a confusor to the point of the example. Additionally, is it maybe the case that the differences are appropriate? E.g. regions are much more likely to be at least partially perturbed in RISE than the sliding-window method.

Purg:
If we're all happy enough with the change in narrative with the current notebook state, I'll squash the change so that there is just one commit (one "update" of the notebook).

brianhhu:
LGTM, feel free to squash if they're aren't any other objections.

Purg:
> LGTM, feel free to squash if they're aren't any other objections.

I just have to now fix my blob size issue, and then I'll be GTG for a merge baring further objection.

Purg:
> I just have to now fix my blob size issue, and then I'll be GTG for a merge baring further objection.

Oh! Right, depending on #9. Will mark as draft until that gets merged. Forgot to do that.

brianhhu (approved):
LGTM!

PR #12, opened 2021-06-03
Update RTD link

brianhhu:
Fix the RTD link so that documentation points to latest instead of stable.


Purg:
LGTM but needs a change note in `docs/release_notes/pending_release.rst`

Purg (approved):

PR #15, opened 2021-06-08
Update perturbation API for streaming iteration

Purg:
This update comes with paired suggestions to our [v0.2 API draft here](https://docs.google.com/document/d/1Er6RQxz2rO5t_yp1LCZP81Cg38GBz0wRD4hfZoiC3Mo/edit#heading=h.u94qan3uz8xk).


brianhhu (commented):
Overall, LGTM! I didn't find any issues with the PR. Assuming the same random seed, we should get the same saliency output regardless of number of threads, correct? Do we also plan to update the example demo notebook to reflect the new streaming interface?

Purg:
Fantastic point, I forgot to update the notebook. Update incoming.

Purg:
> Assuming the same random seed, we should get the same saliency output regardless of number of threads, correct?

If threading is used, order of invocation of the random aspects in the perturbation component is not guaranteed, so the same perturbation output is not guaranteed. Theoretically, the same *number* of invocations will be the same, so I think you are correct in that the *set* of perturbations will be the same. If the saliency map generator is not sensitive to the input order of perturbed parts, then I think the output will be consistent with seeding regardless of threading.

brianhhu:
Great, that tracks my understanding. Comparing the old notebook and the new notebook, the sliding-window based saliency maps look the same (i.e. passes this functional end-to-end test). There are slight differences for the RISE method, but probably understandable since there is some additional stochasticity.

vbhavank (approved):
@Purg Looks great! Thank you.

PR #16, opened 2021-06-08
Implement Similarity Based Saliency Maps

vbhavank:
The following PR brings in a new implementation of Similarity-Based Saliency Maps [[paper](https://openaccess.thecvf.com/content_CVPRW_2019/html/Explainable_AI/Dong_Explainability_for_Content-Based_Image_Retrieval_CVPRW_2019_paper.html)] based on ImageSimilaritySaliencyMapGenerator.


xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
```diff
@@ -0,0 +1,61 @@
...
+
+        if len(perturbed_proximity) != len(perturbed_masks):
+            raise ValueError("Number of perturbation masks and respective",
+                             "confidence lengths do not match.")
+
+        # Iterating through each distance and compare it with
+        # its perturbed twin
+        diff_abs = abs(perturbed_proximity - original_proximity)
+
+        diff = np.transpose(np.clip(diff_abs, a_min=0, a_max=np.max(diff_abs)))
```
brianhhu, line 48: Is a_max necessary here? np.clip documentation says if a_max is None, no clipping is done on the max value.

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
```diff
@@ -0,0 +1,61 @@
+from xaitk_saliency import ImageSimilaritySaliencyMapGenerator
+from xaitk_saliency.utils.masking import weight_regions_by_scalar
+
+import numpy as np
+from sklearn.preprocessing import minmax_scale
+from sklearn.metrics.pairwise import euclidean_distances
```
brianhhu, line 6: [scipy.spatial.distance.cdist ](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.cdist.html#scipy.spatial.distance.cdist) may be slightly more general and future-proof here. It allows for passing in of a metric argument, which can be any distance function, e.g. Euclidean distance, cosine distance, etc.

brianhhu (commented):
In general, LGTM with some minor comments. I would double check the use of absolute value, as I don't think that's correct.

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
```diff
@@ -0,0 +1,61 @@
...
+        # Computing proximity between original image1 and perturbed image2 feature vectors.
+        perturbed_proximity = euclidean_distances(ref_descr_1.reshape(1, -1), perturbed_descrs)[0]
+
+        if len(perturbed_proximity) != len(perturbed_masks):
+            raise ValueError("Number of perturbation masks and respective",
+                             "confidence lengths do not match.")
+
+        # Iterating through each distance and compare it with
+        # its perturbed twin
+        diff_abs = abs(perturbed_proximity - original_proximity)
```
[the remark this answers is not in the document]
vbhavank: I got this absolute value from the previous interface code. After referring to Bo's original code, I removed the absolute. The saliency maps look lesser noisier without the absolute. Thank you for catching this.

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
vbhavank: Changed a_max to None.

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
vbhavank: Great idea, I added support for all distances supported by scipy. Thank you.

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
brianhhu: Really minor point, but I would recommend the following for clarity:

```python
diff = np.transpose(np.clip(diff_abs, a_min=0))
```

We could also potentially rename `diff_abs` for clarity, since we removed the previous `abs` operation.

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
brianhhu: Really minor point, but I would use the following import, and then just use `cdist()` everywhere else:

```python
from scipy.spatial.distance import cdist
```

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
vbhavank: Done.

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
vbhavank: Done.

xaitk_saliency/utils/masking.py
```diff
@@ -104,6 +104,10 @@ def weight_regions_by_scalar(
     :param masks: Mask array in the [nMasks, Height, Weight] shape format.
 
     :return: A numpy array representing the weighted heatmap.
+
+    Note masks can have pixel regions that are never masked raising a
```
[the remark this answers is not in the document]
Purg: If I do this locally I get:
```python
In [6]: import xaitk_saliency.utils.masking, numpy as np

In [7]: scalar_vec = np.array([[0.1,0.9],[0.9,0.1]])

In [8]: m = np.array(
   ...:    [[[0,0,1],
   ...:      [0,0,1],
   ...:      [1,1,1]],
   ...:     [[1,1,1],
   ...:      [1,0,0],
   ...:      [1,0,0]]])

In [9]: xaitk_saliency.utils.masking.weight_regions_by_scalar(scalar_vec, m)
Out[9]: 
array([[[0.1, 0.1, 0. ],
        [0.1, 0.5, 0.9],
        [0. , 0.9, 0.9]],

       [[0.9, 0.9, 0. ],
        [0.9, 0.5, 0.1],
        [0. , 0.1, 0.1]]])
```
which I think is the desired result?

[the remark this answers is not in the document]
Purg: The diff of my local change being
```
diff --git a/xaitk_saliency/utils/masking.py b/xaitk_saliency/utils/masking.py
index 50fc893..ccd44c4 100644
--- a/xaitk_saliency/utils/masking.py
+++ b/xaitk_saliency/utils/masking.py
@@ -57,6 +57,8 @@ def weight_regions_by_scalar(
     sal_across_masks = np.transpose(heatmap.sum(axis=2))
 
     # Compute final saliency map by normalizing with sampling factor.
-    final_heatmap = sal_across_masks/(1 - masks).sum(axis=0)
+    m_sum = (1 - masks).sum(axis=0)
+    m_sum[m_sum == 0] = 1.
+    final_heatmap = sal_across_masks/m_sum
 
     return final_heatmap
```

tests/impls/vis_sal_similarity/test_similarity_scoring.py
```diff
@@ -0,0 +1,107 @@
...
+        assert impl.is_usable() and isinstance(impl, ImageSimilaritySaliencyMapGenerator)
+
+    def test_default_param(self) -> None:
+        """
+        Test default construction.
+        """
+        impl = SimilarityScoring()
+        assert impl.proximity_metric == 'euclidean'
+
+    def test_get_config(self) -> None:
```
Purg, line 27: I would like to point you towards a config test helper in smqtk-core located at `from smqtk_core.configuration import configuration_test_helper`.

Purg: Oh my, I even wrote documentation on this: https://smqtk-core.readthedocs.io/en/latest/plugins_configuration.html#help-with-writing-unit-tests-for-configurable-implementing-types

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
```diff
@@ -0,0 +1,90 @@
...
+        ‘kulsinski’, ‘mahalanobis’, ‘matching’, ‘minkowski’, ‘rogerstanimoto’,
+        ‘russellrao’, ‘seuclidean’, ‘sokalmichener’, ‘sokalsneath’,
+        ‘sqeuclidean’, ‘wminkowski’, ‘yule’.
+    """
+
+    def __init__(
+        self,
+        proximity_metric: str = 'euclidean'
+    ):
+        self.proximity_metric: str = proximity_metric
```
Purg, line 40: Is there any way we can check that this is a valid input at construction time with some referral to the utilized function/module? Worse case we could possibly do a very simple invocation to the function to check that this is a valid parameter for it.

pyproject.toml
```diff
@@ -31,7 +31,7 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = ">=3.7,<3.10"
```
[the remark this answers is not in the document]
vbhavank: I had to change this to add scipy to poetry. Please let me know if there is a better way to check for the minimum version.

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
vbhavank: I created a dummy construction to check if proximity_metric is valid. An alternative would be to store all the supported strings, but I could see that cause tracking issues across versions, so I went with the current fix.

tests/impls/vis_sal_similarity/test_similarity_scoring.py
vbhavank: Sweet! Used the configuration_test_helper. Please let me know what you think

xaitk_saliency/utils/masking.py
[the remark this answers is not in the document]
vbhavank: Yea this is a good fix. I forced values to 1 as per your suggestions. I will be removing the nan_to_num function in *scoring.py if we think this looks good.

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
```diff
@@ -0,0 +1,86 @@
...
+    perturbations (as would be output from a
+    `PerturbImage` implementation.
+
+    The perturbation masks used by the following implementation are
+    expected to be of type integer. Masks containing values of type
+    float are rounded to the nearest value and binarized
+    with value 1 replacing values greater than or equal to half of
+    the maximum value in mask after rounding while 0 replaces the rest.
+
+    param proximity_metric: The type of comparision metric to be used
```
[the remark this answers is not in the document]
vbhavank: Added more docs regarding the limitation around proximity_metric.

docs/release_notes/pending_release.rst
```diff
@@ -24,6 +24,7 @@ Implementations
 * Add new occlusion based classifier scoring in accordance to the v0.1 API draft for ImageClassifierSaliencyMapGenerator.
 
 * Add new RISE based perturbation algorithm in accordance to the v0.1 API draft for PerturbImage
+* Add new similarity based scoring algorithm in accordance to the v0.1 API draft for ImageSimilaritySaliencyMapGenerator
```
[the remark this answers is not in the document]
vbhavank: Thank you, I missed this while updating my branch with master.

.mypy.ini
```diff
@@ -48,3 +48,6 @@ ignore_missing_imports = True
 
 [mypy-tqdm.*]
 ignore_missing_imports = True
+
+[mypy-scipy.*]
```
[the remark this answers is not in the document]
vbhavank: Done.

pyproject.toml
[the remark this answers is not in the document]
Purg: No, this is fine. I was just surprised at `scipy` being that pedantically specific with their python version spec. 3.10 is not even out of beta yet... OK, it is what it is...

tests/impls/vis_sal_similarity/test_similarity_scoring.py
Purg: Use LGTM. I'd say update the name and description (comment) of the test however to reflect the change in what's happening here.

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
```diff
@@ -0,0 +1,98 @@
...
+        # Iterating through each distance and compare it with
+        # its perturbed twin
+        diff = perturbed_proximity - original_proximity
+
+        diff = np.transpose(np.clip(diff, 0, None))
+        # Weighting perturbed regions with respective difference in confidence
+        sal = weight_regions_by_scalar(diff, perturbed_masks)
+
+        # Converting nan values to zero.
+        sal = np.nan_to_num(sal)
```
Purg, line 89: I assume this is the line that may be removed with the update to the `weight_regions_by_scalar` function?

xaitk_saliency/utils/masking.py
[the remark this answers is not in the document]
Purg: `weight_regions_by_scalar` change LGTM.

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
Purg: :+1:

pyproject.toml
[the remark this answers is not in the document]
Purg: My other surprise was that we already had a transitive dependency on scipy version 1.6.1 and had no issue with the `^3.7` python versioning. So this must be something that was introduced in either the `.2` or `.3` patch updates.

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
brianhhu: This will probably also require a separate PR for the OcclusionScoring method as well?

https://github.com/XAITK/xaitk-saliency/blob/master/xaitk_saliency/impls/vis_sal_classifier/occlusion_scoring.py#L48

Purg: Given the scale of where we are, I won't complain with just removing that line in `occlusion_scoring.py` in this PR.

Purg:
That's one flavor. I had in my head the use of 4 unique images, like having a similar cat or something that doesn't match at all. The important bit is that we example something that is informative for the purpose.

tests/impls/vis_sal_similarity/test_similarity_scoring.py
vbhavank: Done. Thank you for the pointer.

xaitk_saliency/impls/vis_sal_similarity/similarityscoring.py
vbhavank: Sounds good, I removed the nan_to_num conversion in both occlusion_scoring.py and similarityscoring.py

vbhavank:
> That's one flavor. I had in my head the use of 4 unique images, like having a similar cat similar to the dual-subject image, or something that doesn't match at all. The important bit is that we example something that is informative for the purpose.

@brianhhu @Purg I added a 4x4 image grid of cross and self-similarity.

vbhavank:
> To add to what @Purg mentioned, we'll also need the license file for the dog image. Without increasing the number of images, we could swap the query and retrieved images (this would provide another example), as well as show self-similarity for each of the two images. If I understand Paul correctly, this would fill out the 2x2 query retrieved pairings.

@Purg Regarding the license for dog.png, does *.txt content count as code duplication?

Purg:
> @Purg Regarding the license for dog.png, does *.txt content count as code duplication?

It's not "code" duplication but it's duplication nonetheless. I believe we discussed offline how I think we can just replace the existing "catdog-LICENSE.txt" and the "new" one with just a "coco-imagery-license.txt" or similar.

Purg:
> @brianhhu @Purg I added a 4x4 image grid of cross and self-similarity.

This presentation in the final cell is still a little uninformative to me, but if @brianhhu is alright with this I can be as well. I don't want to bikeshed on the notebook here so as to get the underlying algo impl merged. We can fiddle more with the notebook in a later PR.

brianhhu:
Yeah, I agree with @Purg that the presentation in the notebook could be improved. We could also strive for consistency with our other notebook (e.g. in terms of helper functions, etc.) However, I don't want to hold this PR up if we're satisfied with the implementation portion itself, we can always edit the notebook in a later PR.

Purg:
Just to circle back, what's pending for this PR still?

My understanding is
- [ ] Replace dog image with one that we have a URL for
- [ ] change license text file to one that is more generically named for included COCO dataset images.

examples/SimilarityScoring.ipynb
```diff
@@ -0,0 +1,215 @@
...
+    "    for i, (img, mask) in enumerate(slid_algo(p_image)):\n",
+    "        pertmask.append(mask)\n",
+    "        pert_img.append(img)\n",
+    "    # Compute feature vectors on perturbed version of the test image\n",
+    "    prtbd_feat = np.array([predict_with_model(model, image_loader(ptrbed)) for ptrbed in pert_img])\n",
+    "    return np.array(pertmask), prtbd_feat\n",
+    "\n",
+    "# Test images containing cat and dog.\n",
+    "test_image1 = Image.open(test_image1_path).resize((imsize, imsize))\n",
+    "test_image2 = Image.open(test_image2_path).resize((imsize, imsize))\n",
```
Purg, line 119: I get errors running this notebook saying that `imsize` is not defined here.

Purg (changes requested):

examples/SimilarityScoring.ipynb
vbhavank: Good catch, corrected it to track `model_input_size`

Purg (approved):
LGTM

brianhhu (approved):
LGTM- per our discussion, none of the issues I raised are blocking this PR. I've created some new issues in our backlog to track them going forward. Thanks!

PR #19, opened 2021-06-14
Sonar properties within secret context

Purg:


Purg:
ideally the SonarCloud job will run here. If so I'm sufficiently convinced that it will "work" on internal branches and master mergers/scheduled runs.

PR #22, opened 2021-06-16
Add support for positive and negative saliency

brianhhu:
Draft PR to update docstrings in interfaces to add support for positive and negative saliency. Update classifier and similarity implementations with new normalization to put values in [-1, 1].


Purg:
Update with rebase on #26 as well as added a notebook update to separately display positive/negative sides of the saliency map returned.

Purg (approved):
LGTM, but then again I did some of the final changes.

xaitk_saliency/impls/vis_sal_classifier/rise_scoring.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    def __init__(
+        self,
+        p1: float = 0.0,
+    ):
+        """
+        Generate RISE-based saliency maps with optional p1 de-biasing.
+
+        :param p1: De-biasing parameter based on the masking probability.
```
brianhhu, line 33: Should we add additional information to say that p1 is bound between [0,1]? I also realized that the RISEPerturbation implementation doesn't explicitly state that this value should be in range [0,1], but maybe that is assumed because it is a probability.

tests/impls/vis_sal_classifier/test_rise_scoring.py
```diff
@@ -0,0 +1,57 @@
...
+from tests import DATA_DIR, EXPECTED_MASKS_4x6
+
+
+class TestRiseScoring:
+
+    def test_configuration(self) -> None:
+        """
+        Test configuration aspects.
+        """
+        inst = RISEScoring(p1=44.777)
```
brianhhu, line 16: Related to my other comment, but p1 I think should probably be bound between [0,1] as it should be the same as the RISE masking probability.

xaitk_saliency/impls/vis_sal_classifier/rise_scoring.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        self.p1 = p1
```
brianhhu, line 35: Would it make sense to add some assert here that p1 is in [0,1]? Similar consideration for RISEPerturbation too.

brianhhu (commented):
I left a few minor comments (mainly about the p1 parameter), but everything seems good to go to me. Minor nitpick: we can fix the OcclusionSaliency Jupyter notebook to use 224x224 for model_input_size in cell 3 (instead of the incorect 224x**244**). We might also want to update the SimilarityScoring notebook at some point. Thanks!

xaitk_saliency/impls/vis_sal_classifier/rise_scoring.py
```diff
@@ -0,0 +1,71 @@
...
+    with value 1 replacing values greater than or equal to half of
+    the maximum value in mask after rounding while 0 replaces the rest.
+    """
+
+    def __init__(
+        self,
+        p1: float = 0.0,
+    ):
+        """
+        Generate RISE-based saliency maps with optional p1 de-biasing.
```
brianhhu, line 31: We can also reference this de-biased RISE version:

```
Based on Hatakeyama et. al: https://openaccess.thecvf.com/content/ACCV2020/papers/Hatakeyama_Visualizing_Color-wise_Saliency_of_Black-Box_Image_Classification_Models_ACCV_2020_paper.pdf
```

xaitk_saliency/impls/vis_sal_classifier/rise_scoring.py
Purg: I'll add a note to both constructor doc-strings.

xaitk_saliency/impls/vis_sal_classifier/rise_scoring.py
Purg: I'll add an exception raise and a unit-test for the check.

xaitk_saliency/impls/vis_sal_classifier/rise_scoring.py
Purg: Added to the class-level doc-string.

Purg:
@brianhhu I believe I have addressed the items that you commented on.

brianhhu:
LGTM! We should create a new PR to update the SimilarityScoring implementation and associated Jupyter notebook. Thanks~

PR #25, opened 2021-06-18
Fix saliency map norm to remove class cross-influence

Purg:


brianhhu (approved):
LGTM, thanks!

examples/.gitignore
```diff
@@ -1,2 +1,2 @@
 **/imagenet_classes.txt*
-
+/*.jpg
```
vbhavank, line 2: No comment, but I am curious about what this was trying to achieve, given my experience with git is limited?

vbhavank (approved):

examples/.gitignore
brianhhu: This will ignore the downloaded *.jpg files in the git repo such that they don't show up as untracked files.

Purg: This is a specific sub-directory scope git ignore file, so this means "ignore any files matching the glob '*.jpg' in this directory."

vbhavank: Got it. Thank you so much for the clarifications. I assumed it was something related to catdog.jpg surgery. I was interested in learning how that fix would look.

PR #28, opened 2021-06-25
Demo use of superpixel masks with OcclusionScoring saliency

brianhhu:
This PR adds a demo Jupyter notebook similar to `OcclusionSaliency.ipynb` that shows the use of arbitrary masks (based on superpixels) for saliency map generation.


Purg:
These different notebooks are all starting to look the same (not a specific criticism of this one, which generally looks fine). We may want to re-organize them to be more focused on what makes them special and highlight at the top of the notebook instead of being nearly a copy-paste of the base occlusion scoring notebook. E.g. this one could maybe highlight that perturbations/masks may originate from outside sources than just `PerturbImage` implementations and the use of skimage quickshift first.

I see that this one applies a little less to my comment on one of the other notebook PRs w.r.t the "app" being repeated, since this one modifies it a little, but it still seems like we could do something there to lessen the repeated content.

Purg (approved):
LGTM. Meta-comment above more applies to the whole suite of notebooks we're adding.

brianhhu:
@Purg I updated the superpixel saliency notebook, swapping around some of the sections, and making mask generation its own section. I also downplayed the "app" portion of the notebook. Let me know if this looks good, and if so, feel free to merge it. Thanks!

PR #29, opened 2021-06-29
Add VIAME demo notebook

brianhhu:
This PR adds a Jupyter notebook demonstrating preliminary integration with VIAME (via bioharn).


examples/VIAME_OcclusionSaliency.ipynb
```diff
@@ -0,0 +1,504 @@
...
+    "*Note*: When using the `from_config_dict` method below,\n",
+    "it is not unusual for the below code to throwback a warning\n",
+    "indicating missing psycopg2 and caffe modules.\n",
+    "Please disregard any such warnings.\n",
+    "\n",
+    "Reference for implementation: Zeiler MD, Fergus R. Visualizing \n",
+    "and understanding convolutional networks. InEuropean conference\n",
+    "on computer vision 2014 Sep 6 (pp. 818-833). Springer, Cham.\n",
+    "\n",
+    "License for test image used may be found in 'COCO-LICENSE.txt'."
```
Purg, line 44: Should we update this line to reference something appropriate for the new imagery?

examples/VIAME_OcclusionSaliency.ipynb
```diff
@@ -0,0 +1,504 @@
...
...qXw8hwnBjFfMOHxy2bJJ9w7a4hjnH2CFnAiM7DB6p+BdEXO5RYF7IqkNBB9yDETocen8W0+rPjFByLltH2Mq/qbZvGvk9+q65JVE3hIs5Xd4M222dmWNvEG8nBVuYSNW2AwWEGgWfHE8BXok4mZO+VQjuwCYLUlMEhT55wfiJdxrXidW76S+p7/w8XlTB9xrovd99jnUDF/uyI2TLG0I9+mBkEddp7Bm9KUQN6GOchhcvbHMqiXJPHw8CBLDYKB085b5yFnaDj3SgTROcv84zjQnbCJOGYHbft3Holq/TN/gp39+Pi40heIY3ZyiRl7jUIVJx2EFsZcjAF5ic9FH4kMZ/JI4F1UMmZ4Rkrp4RS4YkOh5KyGwWLv4l6hODtIHVYWpeCV+jGHYKR+/puXnglC2mxaOB5iRBxY6cdgCnAxguDHyE7ECl7Y1nfhXH5wIg+acQkfMwuDcvq8InfSMh28tbe3t5eXl7///vvp6Un21GTjiDiD4B3hmaoqeZJZARewOS19fHz4AGU4PHA8s8XYRli8W8C9947sSha8TBL3ESErPzGzZQlf/6MV89XYrhYhkZjRgWu7cREwviqbYkhYDawmQ8DAAzyZdIMZrTjHfz/YMhD+8fHx7du3y7hlG13nzQF+mqrIzdWUnZTBvIGq/JyPDwmTFCtKYOZktPDeP188hHObfrKDysA4N3aLI27IAJsJ5X2js4+wZJ/RTRdPnvXaKLSJsNY5r9ehMozgUiFxs39WbqFtxtiG5ZnSWCv3sRwnRYAXbd2GX3qOdKA2zpVKQQwcpt/Oep29IB+jT2gYGfBwHF1ckIjP5ITP/dztDWDSLxQSQ+6rQGxQ7uWJAy/uolg5LuFFPAYltTLMEpLbgITiVtSNJYOpMlmrFso5XZkRMA68NRYj53r/sJMcY6askITUyCBmAErzbs+88q2Uq8t0UVHpqpdOqUQb9hEGLHmwk5UvnBWGf/5JDiHolU3MWD2rrGAXVm2szh3HARv3HEeAsC8hK3Lo3Vmpj1suoTgOyty7RbCXoiy2k9WHfXxQju6+vjcoQKbIBaqdGYEEdgNiSO9sJJVg8nKaN4jSrvROGwFuBE1WQh4weB7p7UWk31Ftqt982Co3JsgHli7+GLZSqvCk8PSFcUqWi9270yp3lGm+zQ6tk0EJBteHeOyLhMN8yDsVkzzcvjpoV0Ua/
```
Purg, line 186: Is there a difference with this function compared to the other notebooks? If not, we may want to break this function out into a helper module here in the `examples` dir for use across the multiple notebooks that are utilizing it.

examples/VIAME_OcclusionSaliency.ipynb
```diff
@@ -0,0 +1,504 @@
...
+    "sorted(dset.boxsize_stats()['perclass'].items(), key=lambda x: x[1]['stats']['shape'][0], reverse=True)[:5]"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# The test image\n",
+    "We will test this application on the following image.\n",
+    "We know that this image contains the ImageNet classes of \"boxer\" and \"tiger cat\"."
```
Purg, line 128: We should update this narrative component to discuss the new image, like what it is called and which of our two classes picked out in out blackbox formulation this is the true class of this image. If we're considering this one image maybe we just want the one, true class back from the blackbox?

docs/release_notes/pending_release.rst
```diff
@@ -20,6 +20,11 @@ Utils
     duplicated across `PerturbImage` implementations. Added both batch and
     streaming versions of this utility.
 
+Documentation
```
Purg, line 23: There is no hard constraint on order of things listed here but my OCD compels me to mention that alphabetically ordered sections is my preference.

examples/VIAME_OcclusionSaliency.ipynb
brianhhu: Checked in with Matt- all images should be publicly available (and available on viame.kitware.com), but NOAA doesn't seem to have the concept of an image license (and it's unclear if asked, whether they will know enough to create the appropriate license). I will try to provide appropriate attribution (sans license).

docs/release_notes/pending_release.rst
brianhhu: I have fixed this now.

examples/VIAME_OcclusionSaliency.ipynb
brianhhu: It's largely the same, but there are small differences in how bioharn/netharn requires inputs for its predict function (i.e. a list of numpy arrays). I would push for getting this notebook up first, and we can discuss whether pulling out that function makes sense in the future.

examples/VIAME_OcclusionSaliency.ipynb
brianhhu: Good catch- I think I'm going to go with the ground-truth class and the predicted class (and try to find an image where these are not the same). Thanks~

examples/VIAME_OcclusionSaliency.ipynb
Purg: agreed :+1:

Purg (approved):
LGTM

PR #31, opened 2021-07-07
Change PIL Image use into numpy ndarray

Purg:


Purg:
And @nitesh201 if you find extra time and want to review something.

docs/index.rst
```diff
@@ -12,6 +12,7 @@ Welcome to xaitk-saliency's documentation!
 
    installation
    releasing
+   design
    miscelaneous
```
brianhhu, line 16: Minor spelling nitpick: `miscelaneous` -> `miscellaneous`.

Might have to update the spelling in the corresponding rst file as well (i.e. miscelaneous.rst).

brianhhu (approved):
This LGTM, and seems to have been a relatively straightforward change since PIL and numpy play pretty well together already. Do we want to cover all the required changes to the example notebooks (e.g. OcclusionSaliency, SimiliarityScoring, and the planned MONAI/VIAME ones) in a separate PR?

docs/index.rst
Purg: Must have missed the spelling squiggles on this. Will correct.

Purg:
> This LGTM, and seems to have been a relatively straightforward change since PIL and numpy play pretty well together already. Do we want to cover all the required changes to the example notebooks (e.g. OcclusionSaliency, SimiliarityScoring, and the planned MONAI/VIAME ones) in a separate PR?

Me missing updating the example notebooks that are current as of this PR's state is a mistake on my part. The other notebooks that are in unmerged PRs will be separately responsible for updates if this PR merges first.

Purg:
@brianhhu I've update the occlusion and similarity notebooks to once again function correctly with the toolkit.

examples/OcclusionSaliency.ipynb
```diff
@@ -90,7 +80,7 @@
     "test_image_filename = 'catdog.jpg'\n",
     "plt.figure(figsize=(12, 8))\n",
     "plt.axis('off')\n",
-    "plt.imshow(PIL.Image.open(test_image_filename))"
+    "_ = plt.imshow(PIL.Image.open(test_image_filename))"
```
brianhhu, line 83: Just out of curiousity- why was this null output change required?

examples/OcclusionSaliency.ipynb
```diff
@@ -124,14 +114,14 @@
...
     "    sal_map_generator: ImageClassifierSaliencyMapGenerator,\n",
     "    fill: Optional[Union[int, Sequence[int]]] = None,\n",
     "    vis_mask_examples: bool = True,\n",
     "):\n",
     "    # Load the image\n",
-    "    ref_image = PIL.Image.open(image_filepath)\n",
+    "    ref_image = np.asarray(PIL.Image.open(image_filepath))\n",
     "    # Use the perterbation API implementation input to generate a bunch of images.\n",
     "    # We will use the outputs here multiple times later so we will just aggregate\n",
     "    # the output here.\n",
```
brianhhu, line 127: We have this line in the notebook (not sure if it's a carry-over from the PIL-based interface):

```
pert_masks = np.asarray(pert_masks)
```

Is this required since I believe the perturber interface should return np.array based masks already?

examples/OcclusionSaliency.ipynb
Purg: While functionally required, I noticed that this will help reduce the diffs made when rerunning the notebook: the return from `plt.imshow` constantly changes (incurring a diff), while the result from plotting is consistent across runs (no diff in image data).

If the last line in a cell has a return, then it is printed: in this case the `plt.imshow` returns an Axes class instance. This object instance `repr()` return includes a memory address, which is of course a different hash nearly every time this is run. By assigning this to a variable, the last line no longer "returns" anything and the instance repr is not printed.

examples/OcclusionSaliency.ipynb
Purg: This is indeed an unnecessary holdover, from the previous streaming interface actually. I'll push a correction momentarily.

examples/OcclusionSaliency.ipynb
brianhhu: Ok, great- thanks for the clarification and that makes a lot of sense. We should feel free to use this trick for future demo notebooks.

vbhavank (approved):
LGTM!

PR #32, opened 2021-07-08
Higher-level image classifier saliency generator interface

Purg:


pyproject.toml
```diff
@@ -37,8 +37,9 @@ numpy = "^1.20.3"
 opencv-python-headless = "^4.5.1"
 scikit-image = "^0.18.1"
 scikit-learn = "^0.24.2"
+# master branch after merge if `ClassifyImage` interface.
+smqtk-classifier = {git = "https://github.com/Kitware/SMQTK-Classifier.git", rev = "aa0dedf7929cbc3a0d71820691ddba450aea4198"}
```
brianhhu, line 41: How stable do we expect `smqtk-classifier` to be? Do we anticipate any breaking changes that might have a downstream effect on `xaitk-saliency`?

xaitk_saliency/impls/vis_sal_image_classifier_blackbox/occlusion_based.py
```diff
@@ -0,0 +1,100 @@
...
+        class_list = blackbox.get_labels()
+        # Input one thing so assume output of one thing.
+        ref_conf_dict = list(blackbox.classify_images([ref_image]))[0]
+        ref_conf_vec = np.asarray([ref_conf_dict[la] for la in class_list])
+        pert_conf_mat = np.empty(
+            (perturbation_masks.shape[0], ref_conf_vec.shape[0]),
+            dtype=ref_conf_vec.dtype
+        )
+        pert_conf_it = blackbox.classify_images(
+            occlude_image_streaming(
```
brianhhu, line 55: Is `blackbox.classify_images` a streaming method? In general, do we expect the `occlude_image_streaming` method to outperform the batch version? I thought that in our latest comparisons, the batch version may be better if we don't return the perturbed images themselves (e.g. sliding window perturbation), but I could be mis-remembering.

xaitk_saliency/interfaces/vis_sal_image_classifier_blackbox.py
```diff
@@ -0,0 +1,99 @@
...
+        blackbox over some image of interest.
+
+        The input reference image is expected to be in matrix form and be in
+        either a `H x W` or `H x W x C` shape format.
+
+        Output saliency map matrix should be in the shape `nClasses x H x W`,
+        be floating-point typed and be composed of values in the `[-1, 1]`
+        range.
+        `nClasses` should be the quantity of unique class labels output by the
+        given classifier blackbox.
```
brianhhu, line 43: Just to check my understanding- this means the blackbox classifier must sub-select class indices (similar to in our current demo notebooks). What are the advantages/disadvantages of this approach vs. passing class indices in the constructor of the `GenerateImageClassifierBlackboxSaliency`? For example, if we wanted to explain new classes, we would need to create a new blackbox classifier instance vs. just changing the class indices param of the constructor.

xaitk_saliency/interfaces/vis_sal_image_classifier_blackbox.py
```diff
@@ -0,0 +1,99 @@
...
+            the quantity of class labels output by the blackbox classifier.
+        """
+        # This is a template method that enforces standard input and result
+        # checks.
+        if ref_image.ndim not in (2, 3):
+            raise ValueError(f"Input image matrix has an unexpected number of dimensions: {ref_image.ndim}")
+        output = self._generate(ref_image, blackbox)
+        # Check that the saliency heatmaps' shape matches the reference image.
+        if output.shape[1:] != ref_image.shape[:2]:
+            raise ShapeMismatchError(
```
brianhhu, line 72: Is this a newly added constraint for the high-level interface (and if so, why)? Using the low-level interface, I think it is possible/allowed for the reference image and the generated saliency maps to be different shapes (you mentioned not leaking image size to the blackbox). Should we make this a constraint throughout our API?

brianhhu (commented):
I added a few comments about design choices, but otherwise it LGTM. I skimmed through the tests, and they seemed reasonable.

pyproject.toml
Purg: Oops! Good catch. This was added to track the branch that was developing the `ImageClassifier` interface before the release so that I could work on this branch. I should make a release of the smqtk-classifier package with the addition of that new interface and then fixup this dependency to reference a published version instead of this hash.

xaitk_saliency/impls/vis_sal_image_classifier_blackbox/occlusion_based.py
Purg: `ImageClassifier.classify_images` is in an iterative/streaming form. So far my testing has shows that the streaming method is strictly faster than the batch mode. I have written and will add a simple benchmarking function to the masking module to actively show this in action going forward (or for someone to point out what I might have done wrong!).

xaitk_saliency/interfaces/vis_sal_image_classifier_blackbox.py
Purg: That was my thought here as I didn't think that it was appropriate for this level of functionality to care about that. If that's a common use-case then we can try to support that better here, however I might argue that such index output selection is more appropriate for the blackbox, or a wrapper thereof, to handle.

xaitk_saliency/interfaces/vis_sal_image_classifier_blackbox.py
Purg: I didn't think this is a new constraint. We previously documented a constraint on `PerturbImage` that output masks are of equivalent HxW to the input reference image. Masks are then input to the generator (`ImageClassifierSaliencyMapGenerator` in this case) that is documented to output masks with equivalent HxW of the masks input. So, by transitivity, the output heatmap is to have the same HxW of the input image.

pyproject.toml
Purg: In terms of "stability" the idea anyway is that this package, currently at least, specifies a min version as well as a pinned version in the `poetry.lock` file for consistent testing / dev-environment. If a new upstream version changes such that xaitk-saliency breaks, then we can more strictly lock-down the acceptable dependency version range. Or we could update xaitk-saliency to use newer versions when it is appropriate for our package to do so.

brianhhu: Sounds good, just wanted to check this since the smqtk-classifier version was pinned to a specific git hash instead of a version number like the other packages.

xaitk_saliency/impls/vis_sal_image_classifier_blackbox/occlusion_based.py
brianhhu: Sounds good, let's leave this as is then- thanks for the clarification.

xaitk_saliency/interfaces/vis_sal_image_classifier_blackbox.py
brianhhu: Yes, you're right- somehow I got confused since I remembered in our example notebooks input images were 480x640, then resized to 224x224 before passing into the black box classifier. I thought the output saliency maps were 224x224, but it looks like they are the same size as the reference input image.

xaitk_saliency/interfaces/vis_sal_image_classifier_blackbox.py
brianhhu: Ok, I think we can revisit this at a later time, but I view the black box as "class-less". The black box could potentially be shared across multiple instantiations of saliency map generation (where the only difference is target class). A wrapper might be a way forward, but you can also see the `target` variable in [Captum's API](https://captum.ai/api/occlusion.html) for reference.

Purg: I'll take a look at the Captum reference and see what might make sense to do here.

brianhhu (approved):
I think you've clarified most of the points I had concerns about, so I'm happy to approve this interface PR. We can also work out any kinks when we start creating implementations.

Purg:
I just want to clean up those LGTM/SonarCloud remarks, and then I'll be good to go

Purg:
@brianhhu Something that I didn't really touch on here is interface naming clarity. It's still seems to me that the names `GenerateImageClassifierBlackboxSaliency` and `ImageClassifierSaliencyMapGenerator` are two similar such that, if I didn't help write the interfaces, I would not know which does which. I think the "lower-level" interface (`ImageClassifierSaliencyMapGenerator`) would make more sense to rename.

How does `GenerateClassifierConfidenceSaliency` sounds? The idea being that this interface deals solely in input confidences that are already computed. I will take alternative suggestions.

Purg:
I also missed adding the impl to the plugin export list, so I'll fix that as well

Purg:
> I also missed adding the impl to the plugin export list, so I'll fix that as well

Fixed.

brianhhu:
> How does `GenerateClassifierConfidenceSaliency` sounds? The idea being that this interface deals solely in input confidences that are already computed. I will take alternative suggestions.

Yeah, I agree the current naming scheme might be confusing. Don't we have a related card on the board this sprint for guidelines on interface/implementation naming? If we go with the `GenerateClassifierConfidenceSaliency` pattern, what would be the equivalent for the renamed similarity and detection interfaces?

brianhhu:
Somewhat related thought, maybe we should change the new interface from `GenerateImageClassifierBlackboxSaliency` to `GenerateImageClassifierSaliency`? This would also require changing the `blackbox` input parameter to `classifier`? This opens the door to at least the possibility of non-black box based saliency methods (which are probably out of scope for our current project, but hopefully implementable by others).

Purg:
> Don't we have a related card on the board this sprint for guidelines on interface/implementation naming?

Yes we do. I was thinking that the renaming of the now-ambiguously-named class would pertain more to this PR, but I can just as easilly slide that change into the branch that provides that documentation + surrounding changes.

Purg:
> Somewhat related thought, maybe we should change the new interface from `GenerateImageClassifierBlackboxSaliency` to `GenerateImageClassifierSaliency`? This would also require changing the `blackbox` input parameter to `classifier`? This opens the door to at least the possibility of non-black box based saliency methods (which are probably out of scope for our current project, but hopefully implementable by others).

I'm a little wary about compounding white- and black-box methods into the same interface. At this level, doing that would cause this interface to be too general: `classifier` would basically become a `void*` outside of the basic API the interface exposes. White box methods are going to require something much more specific than the `ClassifyImage` interface to operate in a stable manner. When supporting white-box methods, the input type ought to formally provide the intrinsic properties that the white-box method(s) require, e.g. the pytorch network module and/or specific layers.

brianhhu:
We discussed this offline, and I think our general conclusion was for keeping `GenerateImageClassifierBlackboxSaliency` for the high-level interface and something like `GenerateClassifierConfidenceSaliency` for the low-level interface.

Purg:
> @Purg and I discussed this offline, and I think our general conclusion was for keeping `GenerateImageClassifierBlackboxSaliency` for the high-level interface and something like `GenerateClassifierConfidenceSaliency` for the low-level interface.

I will use that name change in the upcoming PR that adds interface/class naming rules to the `CONTRIBUTING.md` as well as updates to the appropriate existing interfaces/classes to follow the rule.

PR #35, opened 2021-07-21
fix typo

melster1010:
change "ahold" to "a hold" (also testing first change)

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,7 +1,7 @@
 Installation
 ============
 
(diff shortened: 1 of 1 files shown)


Purg:
Thanks @melster1010! The "failure" here is a pretty benign "lack of a change notes entry" flag. With the more recently added requirement of an approving review, I can probably take the requirement of strict ghostflow passage of "required."

Purg (approved):

PR #37, opened 2021-07-30
Dev/roddy docs

collinsr:
WIP of introduction documentation for interim review.


docs/installation.rst
```diff
@@ -92,7 +92,7 @@ Within the :file:`docs/` directory is a small Python script called
 
 .. prompt:: bash
 
-    python sphinx_server.py
+    poetry run python sphinx_server.py
```
brianhhu, line 95: I haven't checked this locally myself, but I would have assumed that poetry is not required to be able to run the Sphinx server? e.g. if the user had a pip install

docs/introduction.rst
```diff
@@ -0,0 +1,111 @@
+Introduction
+============
+
+The XAITK-Saliency toolkit implements a class of XAI algorithms known
```
brianhhu, line 4: `toolkit` -> `package`? This might help prevent overloading the term "toolkit", and I think we might also refer to `xaitk-saliency` as a package elsewhere (not 100% sure though).

docs/introduction.rst
```diff
@@ -0,0 +1,111 @@
...
+as `saliency algorithms`. The basic application pipeline is shown in Figure 1:
```
brianhhu, line 5: Maybe change to "machine learning application pipeline"?

docs/introduction.rst
```diff
@@ -0,0 +1,111 @@
...
+
+.. figure:: figures/intro-fig-01.png
+
+*Figure 1: A basic AI pipeline.*
+
+In this scenario, an AI algorithm operates on an input (text, image,
+etc.) to produce some sort of output. Saliency algorithms build on
```
brianhhu, line 12: maybe provide examples of output as well?

`output (classification, detection, etc.)`

docs/introduction.rst
```diff
@@ -0,0 +1,111 @@
...
+this as shown in Figure 2:
```
brianhhu, line 13: Linking to the idea of saliency maps introduced in the next paragraph:

"to produce visual explanations in the form of saliency maps as shown in Figure 2 ".

docs/introduction.rst
```diff
@@ -0,0 +1,111 @@
...
+.. note:: The XAITK-Saliency toolkit focuses on providing saliency
+          maps for images.
+
+Image Saliency Maps: An Intuitive Introduction
+----------------------------------------
+
+Figure 4 shows a deep learning pipeline for recognizing objects in
+images; pixels in the image (in green) are processed by the pipeline
+to produce the output (in orange). Here, the system has been trained
+to recognize 1000 object types. Its output is a list of 1000 numbers,
```
brianhhu, line 39: `types` -> `categories`, just to be slightly more consistent with the literature.

docs/introduction.rst
```diff
@@ -0,0 +1,111 @@
...
+.. figure:: figures/cnn-tagged-scaled.png
+
+*Figure 4: Typical object recognition CNN architecture; the image
+(green, left) is processed left to right through the CNN to produce an
+output vector (orange, right). Diagram from* `Krizhevsky et
+al. <https://proceedings.neurips.cc/paper/4824-imagenet-classification-with-deep-convolutional-neural-networks.pdf>`_
+
+
+This operation is "all or nothing" at both ends: the entire image must
+be processed, and the the entire vector must be output. There is no
```
brianhhu, line 53: The motivation for this paragraph is a bit unclear- is this mainly harping on the "black box" nature of these types of systems?

docs/introduction.rst
```diff
@@ -0,0 +1,111 @@
...
+* The input image contains a dog and a cat, and the system gave a high
+  likelihood for "cat" but not "dog". How will the system respond if
+  the cat is removed?
+
+At some level, these questions require a degree of *introspection*;
+the system must produce not only the output, but also some information
+about **how** the output was produced. There are two popular
+approaches to this:
+
+1) The **white box** approach: the system is altered to open or expose
```
brianhhu, line 76: I might couch it this way (needs some polishing): "white box" methods assume access to the internals of the model, including model weights and/or gradients with respect to some loss function (this may be too technical for the lay audience). We can typically perform forward and backward passes through the model.

docs/introduction.rst
```diff
@@ -0,0 +1,111 @@
...
+Requires modifying the model implementation to gain access; may
+require updating as the implementation evolves.*
+
+Black Box Methods
+^^^^^^^^^^^^^^^^^
+*Explanation options are correlated to how the related images are
+generated. pros: indpendent of the model; operates across all models; does not
+require access to the model implementation. cons: requires extra work
+to generate and process the related images; makes only indirect /
+differential observations about the original input / output pair.*
```
brianhhu, line 105: Another con: generally slower/more costly than white-box methods.

brianhhu (approved):

docs/installation.rst
Purg: Use of poetry here is not *required* per se, but it more of a "this should just work" suggestion given that we are assuming poetry for build/environment management for developers. A developer *can* of course create their own virtualenv/conda environment, but comes with assumed environment setup that is out of scope to detail here.

Purg: If it wasn't clear above, I support Roddy's fix here.

docs/introduction.rst
Purg: I assume we don't want to use "black-box" here because that term is introduced later. Maybe we want to more phrase this in terms of different target audiences not being able to understand the "why" of the results from just the input, the outputs and the raw algorithm, like the following questions you pose in bullet points.

docs/introduction.rst
Purg: Even "forward and backward passes" is specific to neural networks. If we want to be more broad about it, we're referring here to when we are able to access the internal workings or parameters of a model such that more algorithm/model-specific introspection may be performed, which can work to produce more detailed explanations for incrementally more targeted audiences, or to make more specific/detailed explanations.

docs/introduction.rst
```diff
@@ -0,0 +1,111 @@
...
+
+1) The **white box** approach: the system is altered to open or expose
+   the internal state of the model. This state is examined while the
+   system is generating the output.
+
+2) The **black box** approach: the model is not opened; instead, we
+   create a series of *related images* which perturb or change the
+   original image in some way. By comparing the original output to the
+   output for the related images, we can deduce certain aspects of the
+   model's behavior.
```
Purg, line 84: Do want our verbiage to be specific to images? If not we might want to change the use of "image" here to a more generic "input."

Purg:
Also see the ghostflow output: whitespace and lack-of-change-note "errors."

docs/introduction.rst
collinsr: sounds good

docs/introduction.rst
collinsr: sounds good

docs/introduction.rst
collinsr: sounds good

docs/introduction.rst
collinsr: sounds good

docs/introduction.rst
collinsr: Yes. Good catch.

docs/introduction.rst
collinsr: Yes, although not so much black-box / white-box but rather the monolithicity of a CNN independent of XAI.

docs/introduction.rst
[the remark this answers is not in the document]
collinsr: That's probably a good idea to have somewhere, although I'm not sure here is the place? Should there be a link to some other resource on xaitk.org ?

docs/introduction.rst
collinsr: noted!

docs/introduction.rst
collinsr: I was going to defer specifics about e.g. gradients, forward/backward passes, etc to the specific sections, and try to leave this text general.

docs/introduction.rst
```diff
@@ -0,0 +1,111 @@
...
+etc.) to produce some sort of output. Saliency algorithms build on
+this as shown in Figure 2:
+
+.. figure:: figures/intro-fig-02.png
+
+*Figure 2: The AI pipeline augmented with a saliency algorithm.*
+
+At a high level, saliency maps are typically a copy of the input with
+regions highlighted to indicate areas that are somehow significant to
+the AI. Figure 3 shows sample saliency maps for text and images.
```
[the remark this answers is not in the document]
collinsr: Hmm. Obviously at the code level, saliency maps exist independently of the source image. But how often does the user see the heatmap without the image? I tried the applied-to approach; how does that read?

docs/introduction.rst
collinsr: sounds good

docs/introduction.rst
```diff
@@ -0,0 +1,111 @@
...
+
+.. figure:: figures/intro-fig-03.png
+
+*Figure 3: Sample saliency maps for (left) text (from* `Tuckey et al.
+<https://arxiv.org/abs/1907.05664>`_) *and (right) images (from* `Dong et
+al. <https://openaccess.thecvf.com/content_CVPRW_2019/html/Explainable_AI/Dong_Explainability_for_Content-Based_Image_Retrieval_CVPRW_2019_paper.html>`_
+*)*
+
+.. note:: The XAITK-Saliency toolkit focuses on providing saliency
+          maps for images.
```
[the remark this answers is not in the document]
collinsr: Sounds good

docs/introduction.rst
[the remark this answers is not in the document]
Purg: :+1:

docs/introduction.rst
Purg: @brianhhu do we have any more specific feedback for this or can this be the way it is for this PR?

docs/introduction.rst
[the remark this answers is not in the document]
Purg: I think we discussed that this level would be something for the xaitk.org content. Is that an accurate understanding?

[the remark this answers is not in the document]
brianhhu: Yes, that's fine with me- we can add a pointer to xaitk.org content (once there is more information about XAI in general on the website). Thanks!

docs/introduction.rst
brianhhu: I think this is fine for now, we can always choose to revisit this after the other sections are completed.

Purg:
Dear @collinsr,

I broke it by messing up an attempt at resolving a file conflict via github's interface. Please forgive me.

PR #38, opened 2021-07-30
Add RISE high-level impl convenience

Purg:


xaitk_saliency/impls/gen_image_classifier_blackbox_sal/rise.py
```diff
@@ -0,0 +1,62 @@
...
+        self,
+        n: int,
+        s: int,
+        p1: float,
+        seed: Optional[int] = None,
+        threads: int = 0
+    ):
+        self._po = PerturbationOcclusion(
+            RISEGrid(n=n, s=s, p1=p1, seed=seed, threads=threads),
+            RISEScoring(p1=p1),
```
brianhhu, line 51: Currently, this provides the de-biased form of RISE, i.e. p1 is coupled between the `RISEGrid` and `RISEScoring` methods. Should we add a flag in the constructor for something like `debiased = True`? This could allow `p1` to be set to 0, which is for standard RISE.

brianhhu (approved):
In general, this looks good to me. As a sanity check, we should probably perform a functional test to make sure the high-level interface produces the same saliency maps as those we have in our example notebooks. Or is there a unit test that we can create for this?

Not necessarily related to this specific PR, but I have questions regarding the consistency of the current naming of files/classes in `xaitk-saliency`. I'll raise a separate issue for this.

xaitk_saliency/impls/gen_image_classifier_blackbox_sal/rise.py
Purg: I thought about it, but decided to keep it simple to start and see if there was any commentary on it! :) I can make it a boolean enable flag. I was thinking to default it to `True` unless you think otherwise.

PR #42, opened 2021-08-03
Periodic dep lock update because we're a library

Purg:
Manual ping to update locked dependencies. Being a library, xaitk-saliency should strive to be compatible with its stated abstract requirements when installed in the wild. Since developers and CI use our checked in, locked updates for consistency, we should check that updating these periodically does not break anything. Breakage may either lead to a constraint being added to the abstract dependencies specified or an issue in need of fixing.

Maybe this should be incorporated into CI somewhere/how, but I don't have an immediate vision on that right now.

### DIFF
--- a/docs/design.rst
+++ b/docs/design.rst
@@ -1,14 +1,44 @@
 Design Decisions
 ================
 
+Concrete Dependencies and Updating
+----------------------------------
+This package, like many others, has dependencies that need to be fulfilled for
+proper functionality to be satisfied.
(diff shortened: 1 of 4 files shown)


brianhhu (approved):
LGTM, thanks! As discussed, we should probably come up with guidance on a cadence for updating this (possibly with each release to start with).

Purg:
@brianhhu about to push an update that adds documentation to the "design decisions" page (couldn't think of a better place).

brianhhu:
LGTM, thanks for adding the extra documentation.

PR #43, opened 2021-08-04
Add sliding window high level impl convenience

Purg:
Adding another higher level implementation "convenience" for the sliding-window/difference-occlusion algorithms. This follows the same encapsulation method as the other RISE convenience implementation.

### DIFF
--- a/docs/release_notes/pending_release.rst
+++ b/docs/release_notes/pending_release.rst
@@ -51,9 +51,13 @@ Implementations
 * Add ``RISEScoring`` implementation, with the ability to also compute a
   de-biased form of RISE with an optional input parameter.
 
-* Add ``RISEStack`` implementation of ``GenerateImageClassifierBlackboxSaliency``
+* Add ``RISEStack`` implementation of the ``GenerateImageClassifierBlackboxSaliency``
   interface as a simple way to invoke the combination of RISE component
   algorithms.
+
+* Add ``SlidingWindowStack`` implementation of the ``GenerateImageClassifierBlackboxSaliency``
+  interface as a simple way to invoke the combination of the Sliding Window
+  perturbation method and the occlusion-based based scoring method.
 
 Misc.
 
--- a/xaitk_saliency/impls/gen_image_classifier_blackbox_sal/slidingwindow.py
+++ b/xaitk_saliency/impls/gen_image_classifier_blackbox_sal/slidingwindow.py
@@ -1,0 +1,57 @@
+from typing import Tuple, Dict, Any
+
+import numpy as np
+from smqtk_classifier import ClassifyImage
+
+from xaitk_saliency.impls.gen_classifier_conf_sal.occlusion_scoring import OcclusionScoring
+from xaitk_saliency.impls.gen_image_classifier_blackbox_sal.occlusion_based import PerturbationOcclusion
+from xaitk_saliency.impls.perturb_image.sliding_window import SlidingWindow
+from xaitk_saliency.interfaces.gen_image_classifier_blackbox_sal import GenerateImageClassifierBlackboxSaliency
+
+
+class SlidingWindowStack (GenerateImageClassifierBlackboxSaliency):
+    """
+    Encapsulation of the perturbation-occlusion method using specifically
+    sliding windows and the occlusion-scoring method.
+
+    :param window_size: The block window size as a tuple with format
+        `(height, width)`.
+    :param stride: The sliding window striding step as a tuple with format
+        `(height_step, width_step)`.
+    :param threads: Optional number threads to use to enable parallelism in
+        applying perturbation masks to an input image. If 0, a negative value,
+        or `None`, work will be performed on the main-thread in-line.
+    """
+
+    def __init__(
+        self,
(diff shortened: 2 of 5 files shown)


docs/release_notes/pending_release.rst
```diff
@@ -51,10 +51,14 @@ Implementations
...
   de-biased form of RISE with an optional input parameter.
 
-* Add ``RISEStack`` implementation of ``GenerateImageClassifierBlackboxSaliency``
+* Add ``RISEStack`` implementation of the ``GenerateImageClassifierBlackboxSaliency``
   interface as a simple way to invoke the combination of RISE component
   algorithms.
 
+* Add ``SlidingWindowStack`` implementation of the ``GenerateImageClassifierBlackboxSaliency``
+  interface as a simple way to invoke the combination of the Sliding Window
+  perturbation method and the occlusion-difference based scoring method.
```
brianhhu, line 60: "occlusion-difference based scoring method" -> "occlusion-based scoring method" seems slightly less confusing.

xaitk_saliency/impls/gen_image_classifier_blackbox_sal/slidingwindow.py
```diff
@@ -0,0 +1,47 @@
...
+from xaitk_saliency.impls.gen_image_classifier_blackbox_sal.occlusion_based import PerturbationOcclusion
+from xaitk_saliency.impls.perturb_image.sliding_window import SlidingWindow
+from xaitk_saliency.interfaces.gen_image_classifier_blackbox_sal import GenerateImageClassifierBlackboxSaliency
+
+
+class SlidingWindowStack (GenerateImageClassifierBlackboxSaliency):
+    """
+    Encapsulation of the perturbation-occlusion method using specifically
+    sliding windows and the occlusion-scoring method.
+    """
```
brianhhu, line 16: In our corresponding RISEStack method, we duplicated the docstring with information on the input parameters. However, I see that's this is missing here (e.g. info on window_size, stride, threads).

xaitk_saliency/impls/gen_image_classifier_blackbox_sal/slidingwindow.py
```diff
@@ -0,0 +1,47 @@
...
+            threads=threads,
+        )
+
+    def _generate(self, ref_image: np.ndarray, blackbox: ClassifyImage) -> np.ndarray:
+        return self._po.generate(ref_image, blackbox)
+
+    @classmethod
+    def get_default_config(cls) -> Dict[str, Any]:
+        # Minor override following SlidingWindow impl because of list/tuple
+        # equality check differences.
```
brianhhu, line 39: Can you unpack this a bit more? Under what scenarios is this `@classmethod`  pattern required?

xaitk_saliency/impls/gen_image_classifier_blackbox_sal/slidingwindow.py
```diff
@@ -0,0 +1,47 @@
...
+        cfg = super().get_default_config()
+        cfg.update(SlidingWindow.get_default_config())
+        return cfg
+
+    def get_config(self) -> Dict[str, Any]:
+        c = self._po.perturber.get_config()
+        c['threads'] = self._po.threads
```
brianhhu, line 46: Why do we have to explicitly set `threads` in the config here, but not in the corresponding RISEStack method?

brianhhu (approved):
LGTM, added some relatively minor review comments, mostly for clarification.

docs/release_notes/pending_release.rst
Purg: Done.

xaitk_saliency/impls/gen_image_classifier_blackbox_sal/slidingwindow.py
Purg: I indeed missed getting the parameter docs here. I will add them.

xaitk_saliency/impls/gen_image_classifier_blackbox_sal/slidingwindow.py
Purg: I thought I would be "smart" and make user of the currying that the `SlidingWindow` implementation does, but I think its both less clear as well as "wrong" (uses defaults of `SlidingWindow`s constructor, which may be different than here).

xaitk_saliency/impls/gen_image_classifier_blackbox_sal/slidingwindow.py
Purg: This has to do with what parameters are defined in the `perturber` instance that are also "shared" with this implementation. In the RISE case, the RISE perturbation implementation had a threading parameter that is set in the `RISEStack` constructor to both that impl *and* the `PerturbationOcclusion` constructor. In this case, with `SlidingWindowStack`, the perturber being drawn from does *not* have a threading parameter, which is only encoded in the `PerturbationOcclusion` instance.

Purg:
@brianhhu PTAL at the latest updates here.

brianhhu:
LGTM- thanks!

PR #50, opened 2021-08-23
Add initial FAQ documentation file

Purg:


brianhhu (approved):
LGTM, similar to the review process document, we should consider if we want to include it in the index so that it shows up in ReadtheDocs. Thanks!

Purg:
> LGTM, similar to the review process document, we should consider if we want to include it in the index so that it shows up in ReadtheDocs. Thanks!

Good catch, let me get it in there.

PR #52, opened 2021-09-06
Remove use of TestCase base-class

Purg:
The baseclass was not utilized in any meaningful way that is not already
handled by PyTest.


brianhhu (approved):
LGTM, thanks!

PR #53, opened 2021-09-13
Consolidate some deps that are not used

Purg:
Also clean-up mypy config a little to also remove ignores for packages
that are not used in the tested scope.


brianhhu (approved):
LGTM

PR #54, opened 2021-09-14
Add initial documentation for implementations

schencej:


docs/implementations.rst
```diff
@@ -0,0 +1,72 @@
+=========================
+Saliency Implementations
+=========================
+
+Included with this toolkit are a number of implementations for the interfaces described in the previous section. Like those interfaces, these initial implementations focus of black-box visual saliency and are separated by use-case. Each implementation provides functionality for a different perturbation or heatmap generation algorithm specific to that use-case.
```
Purg, line 5: I would recommend that each sentence be on a separate line, and use a 80-120 character width maximum.
This makes text blocks like this a bit easier to read in simpler text editors.
This also aids in simplifying change diffs: a spelling or sentence correction does not mark the whole "paragraph" as changed.

E.g. for this paragraph:
```suggestion
Included with this toolkit are a number of implementations for the interfaces
described in the previous section.
Like those interfaces, these initial implementations focus of black-box visual
saliency and are separated by use-case.
Each implementation provides functionality for a different perturbation or
heatmap generation algorithm specific to that use-case.use-case.
```

docs/conf.py
```diff
@@ -47,6 +47,7 @@
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = ["_build", "Thumbs.db", ".DS_Store"]
 
+autoclass_content = 'both'
```
Purg, line 50: What is this property setting for?

brianhhu, line 50: Can we add a short in-line comment for this? That way we can go back and see why this choice was made without having to Google. Thanks!

xaitk_saliency/impls/perturb_image/rise.py
```diff
@@ -49,6 +49,7 @@ def __init__(
     ):
         """
         Generate a set of random binary masks
+
```
brianhhu, line 52: Looks like this newline has to be removed for linting purposes.

docs/implementations.rst
```diff
@@ -0,0 +1,72 @@
...
+.. autoclass:: xaitk_saliency.impls.gen_image_classifier_blackbox_sal.slidingwindow.SlidingWindowStack
+  :members:
+  
+-----------
+Similarity
+-----------
+
+Class: SimilarityScoring
+-------------------------
+.. autoclass:: xaitk_saliency.impls.gen_descriptor_sim_sal.similarity_scoring.SimilarityScoring
```
brianhhu, line 60: Need to check the `proxity_metric` param- it doesn't seem to render properly in ReadTheDocs.

docs/implementations.rst
```diff
@@ -0,0 +1,72 @@
...
+  :members:
+  
+----------
+Detection
+----------
+
+Class: DRISEScoring
+--------------------
+.. autoclass:: xaitk_saliency.impls.gen_detector_prop_sal.drise_scoring.DRISEScoring
```
brianhhu, line 69: Need to check DRISEScoring's `generate` function params- these also do not seem to render properly in ReadTheDocs.

docs/implementations.rst
```diff
@@ -0,0 +1,72 @@
...
+  :members:
```
brianhhu, line 70: Similarly, the `iou` function's `args` and `return` fields could use some work to render properly in ReadTheDocs. Thanks!

Purg:
The DRISE stuff isn't rendering "correctly" because it was written without using the consistent RST-style fieldlists ([section in sphinx docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#field-lists)). Also `DRISEScoring.generate` should not overwrite the doc-string that would be "inherited" from the parent interface function: Doc-strings should generally only be added for novel member functions that have no parent to draw from.

I also noticed another instance of the annoyance that is the parent `Pluggable` interface overriding `__new__` for some checks. I've been teetering on whether the check that is being placed there is really necessary (prevents construction if the classmethod `is_usable` returns False), but it for some reason I don't understand yet interrupts constructor signature introspection, which is why all the docs for __init__ functions here are showing `(*args: Any, **kwargs: Any)`, which is super unhelpful.

brianhhu (commented):
Looks pretty close to me- I made some minor comments about source code changes to get the ReadTheDocs to render properly. Also note that our CI requires adding an entry to the docs/release_notes/pending_release.rst file (under the "Documentation" header).

xaitk_saliency/impls/perturb_image/rise.py
Purg: It's not that there's a new-line here, but that the empty line has trailing white-space. Removing that should resolve the issue.

NOTE: these ghostflow checks about white-space are on a PER-COMMIT basis, so fixing this will require a fixup-commit into the offending commit.

docs/implementations.rst
schencej: Is it okay to do this with a fixup commit?

schencej:
There's currently no documentation for the `DRISEScoring.__init__` function's one parameter

Purg:
> There's currently no documentation for the `DRISEScoring.__init__` function's one parameter

Good catch! We should add it here if we can. Make something like "String label of the distance metric to use. See `scipy.spatial.distance.cdist()` and it's `metric` parameter for more details and possible values."

xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py
```diff
@@ -39,14 +39,14 @@ def __init__(
...
-            A ∩ B / A ∪ B = A ∩ B / (area(A) + area(B) - A ∩ B)
-        Args:
-            box_a: (np.array) bounding boxes, Shape: [A,4]
-            box_b: (np.array) bounding boxes, Shape: [B,4]
-        Return:
-            iou: (np.array), Shape: [A,B].
+        """
+        | Compute the intersection over union (IoU) of two sets of boxes.
+        | E.g.:
+        |    A ∩ B / A ∪ B = A ∩ B / (area(A) + area(B) - A ∩ B)
```
Purg, line 45: We can probably dis-include the title sentence of the doc-string from the formatted block
```suggestion
        Compute the intersection over union (IoU) of two sets of boxes.

        | E.g.:
        |    A ∩ B / A ∪ B = A ∩ B / (area(A) + area(B) - A ∩ B)
```

docs/implementations.rst
```diff
@@ -0,0 +1,75 @@
+=========================
+Saliency Implementations
+=========================
+
+Included with this toolkit are a number of implementations for the interfaces
+described in the previous section.
+Like those interfaces, these initial implementations focus of black-box visual
+saliency and are separated by use-case.
+Each implementation provides functionality for a different perturbation or
+heatmap generation algorithm specific to that use-case.
```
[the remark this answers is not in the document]
schencej: I was trying to get at the fact that they are inherently divided into use-cases because the interfaces that they implement are. But I guess that's pretty obvious and what you wrote portrays the intent of the implementations more clearly. Thanks, I'll change it.

brianhhu (approved):
LGTM. @Purg Please take a final pass through as well. Thanks!

Purg (approved):
LGTM

PR #56, opened 2021-09-16
Add expansion in docs and unittests for "fill" with an image

Purg:
We discussed that a leading "occlusion" method is to blend with blurred versions of the reference image instead of using a constant color fill. It turns out that our occlusion methods already supported this functionally but not in type annotations, documentation or unit tests, all of which have been updated to describe and test use with images.

I have also offline tested the use of this feature with the "OcclusionSaliency" notebook to see how it works with something not-a-unittest and the results were roughly comparable with the base-line.

### DIFF
--- a/docs/release_notes/pending_release.rst
+++ b/docs/release_notes/pending_release.rst
@@ -30,6 +30,11 @@ Tests
 * Removed use of `unittest.TestCase` as it is not utilized directly in any way
   that PyTest does not provide.
 
+Utilities
+
+* Add type annotation, documentation and unit-tests for using image matrices as
+  the fill option instead of just a solid color.
+
 
 Fixes
 -----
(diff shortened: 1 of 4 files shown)


brianhhu (approved):
LGTM- thanks!

Out of scope for the current PR, but any thoughts on how to easily facilitate automatic creation of the fill image (i.e. a blurred version of the `ref_image` with some parameter `sigma`)? Thinking mainly about how to implement this similar to our other "stacked" implementations.

Purg:
> Out of scope for the current PR, but any thoughts on how to easily facilitate automatic creation of the fill image (i.e. a blurred version of the `ref_image` with some parameter `sigma`)? Thinking mainly about how to implement this similar to our other "stacked" implementations.

This maybe addresses what made me "uncomfortable" with the current `PerturbOcclude` implementation (and sub-classes) that exposes a `fill` attribute for a user to maybe know to fill in. Instead, if filling with a constant color is just generally ill-advised, move some hard logic into that base implementation that blurs the input reference image at runtime, taking the radius parameter as a constructor argument (and thus configuration option).

Purg:
@brianhhu I will merge when the CI is green.

PR #57, opened 2021-09-17
Add scikit example notebook

schencej:


examples/Scikit.ipynb
```diff
@@ -0,0 +1,330 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "888638fa",
+   "metadata": {},
+   "source": [
+    "## Introduction\n",
+    "\n",
+    "The purpose of this example is to demonstrate the API's ability to work with a variety of machine learning libraries.\n",
+    "Here we use scikit learn's `LogisticRegression` class to implement a logistic regression model for image classification of the MNIST dataset.\n",
```
brianhhu, line 11:
```suggestion
    "Here we use scikit-learn's `LogisticRegression` class to implement a logistic regression model for image classification of the MNIST dataset.\n",
```

examples/Scikit.ipynb
```diff
@@ -0,0 +1,330 @@
...
+   "id": "08fb1a3e",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from smqtk_classifier import ClassifyImage\n",
+    "from xaitk_saliency import GenerateImageClassifierBlackboxSaliency\n",
+    "from functools import reduce\n",
+    "from math import ceil\n",
+    "\n",
+    "def app(\n",
```
brianhhu, line 179: I think here it would be nice for the app to display the digit below each saliency map, similar to above for the scikit-learn based visualization.

examples/Scikit.ipynb
```diff
@@ -0,0 +1,330 @@
...
+    "    images: np.ndarray,\n",
+    "    image_classifier: ClassifyImage,\n",
+    "    saliency_generator: GenerateImageClassifierBlackboxSaliency\n",
+    "):\n",
+    "    num_imgs = images.shape[0]\n",
+    "    \n",
+    "    # Generate saliency maps\n",
+    "    sal_maps = []\n",
+    "    for i in range(num_imgs):    \n",
```
brianhhu, line 188: I think here, it would be nice to also visualize saliency maps for individual MNIST examples. We could make the story to be: 1) as a form of "local" explanation, show saliency maps for example digit exemplars, and 2) as a form of "global" explanation, show average saliency maps for each digit class.

Related to this, it is not surprising to me that the saliency maps agree with the coefficients learned by the logistic regression model since we are operating in the linear regime (i.e. each coefficient is a measure of influence on the model's final prediction, and since each coefficient maps to a certain pixel, this by definition maps directly to a pixel's influence on the final model prediction).

One thing that could be potentially interesting to try would be to borrow the MLP from the second scikit-learn [example](https://scikit-learn.org/stable/auto_examples/neural_networks/plot_mnist_filters.html#sphx-glr-auto-examples-neural-networks-plot-mnist-filters-py) (which is more than 1 layer) and apply our saliency maps there. The model is more complex and also achieves higher accuracy. I believe in that example they visualize the weights of the first layer, which are not great. This would be a slightly more involved example, but I still think instructive for the reader. @Purg Any thoughts on this?

examples/Scikit.ipynb
```diff
@@ -0,0 +1,330 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e6d18472",
+   "metadata": {},
+   "source": [
+    "## Blackbox Classifier\n",
+    "\n",
+    "Here we wrap our logistic regression in a `ClassifyImage` class to comply with the API's interface."
```
brianhhu, line 211: Might be useful to mention this as part of `smqtk-classifier`:

```suggestion
    "Here we wrap our logistic regression in smqtk-classifier's `ClassifyImage` class to comply with the API's interface."
```

brianhhu (commented):
In general, LGTM- made a few minor comments here and there. I also made some comments about going a bit deeper into the MNIST problem pending Paul's thoughts/feedback.

examples/Scikit.ipynb
```diff
@@ -0,0 +1,330 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "id": "99c51510",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from xaitk_saliency.impls.gen_image_classifier_blackbox_sal.rise import RISEStack\n",
```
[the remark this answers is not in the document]
schencej: Oops... I played around with the RISE algorithm for a while but the maps never looked very good. I'm not sure it's suitable for small images like what MNIST consists of. I forgot to remove the import here.

examples/Scikit.ipynb
```diff
@@ -0,0 +1,330 @@
...
+   "cell_type": "markdown",
+   "id": "888638fa",
+   "metadata": {},
+   "source": [
+    "## Introduction\n",
+    "\n",
+    "The purpose of this example is to demonstrate the API's ability to work with a variety of machine learning libraries.\n",
+    "Here we use scikit learn's `LogisticRegression` class to implement a logistic regression model for image classification of the MNIST dataset.\n",
+    "This is inspired and partially taken from thie example of scikit's website:<br>\n",
+    "&emsp;&emsp;https://scikit-learn.org/stable/auto_examples/linear_model/plot_sparse_logistic_regression_mnist.html\n"
```
Purg, line 13: I think it would be good to add more to this introduction, specifically about what we are adding to this example on top of the reference example from within scikit-learn.

For example, the example shows, effectively, an attention visualization, as we are duplicating here. The visualization of saliency heatmaps below seems very similar. I believe it is insinuated in the "Calling the Application" section, but we could be explicit here stating that the Scikit-Learn example visualization relies on *intrinsics* of the model in order to determine the "attention" visualization (whitebox), however our API's approach is classifier agnostic, "blackbox", and shows very comparable results without knowing anything about the specific classifier model being used, though likely some amount more processing intensive.

examples/Scikit.ipynb
```diff
@@ -0,0 +1,330 @@
...
+    "        return [i for i in range(10)]\n",
+    "    \n",
+    "    def classify_images(self, image_iter):\n",
+    "        \n",
+    "        # Convert image set back to column vector form\n",
+    "        images = np.array([]).reshape(0, 28*28)\n",
+    "        for img in image_iter:\n",
+    "            images = np.vstack((images, img.reshape(1, -1)))\n",
+    "        \n",
+    "        return clf.decision_function(images)\n",
```
Purg, line 232: This technically does not fit the output data type of the interface API (returns iterator of dict instances). This seems to happen to work due to the classes being integers and the output elements index-able arrays. Maybe we don't need to change the code, but at least comment that the API is not being "met" due to special circumstances.

examples/Scikit.ipynb
```diff
@@ -0,0 +1,330 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "dcdab7a0",
+   "metadata": {},
+   "source": [
+    "## Heatmap Generation\n",
+    "\n",
+    "We create a `SlidingWindowStack` object to carry out our image perturbation and heatmap generation."
```
Purg, line 248:
```suggestion
    "We create an instance of `SlidingWindowStack`, an implementation of the `GenerateImageClassifierBlackboxSaliency` interface, to carry out our image perturbation and heatmap generation."
```

examples/Scikit.ipynb
```diff
@@ -0,0 +1,330 @@
...
+    "    \n",
+    "    # Combine respective class maps\n",
+    "    combined_maps = reduce(np.add, sal_maps)\n",
+    "    \n",
+    "    # Plot combined maps\n",
+    "    plt.figure(figsize=(10, 5))\n",
+    "    for i in range(num_imgs):\n",
+    "        max_val = combined_maps[i].max()\n",
+    "        plt.subplot(2, ceil(num_imgs/2), i + 1)\n",
+    "        plt.imshow(combined_maps[i], cmap=plt.cm.RdBu, vmin=-max_val, vmax=max_val)\n",
```
Purg, line 200: I'm not sure the use of `vmin=-max_val, vmax=max_val` is appropriate here. The value is not symmetric in the heatmaps. The scikit-learn example uses this because it takes the global coef matrix absolute-value max.

We also know via API documentation that saliency heatmaps are constrained to return in the `[-1,1]` range. We may want to normalize the `combined_maps` by `len(sal_maps)` and then pass `-1, 1` to `vmin, vmax` respectively. @brianhhu please double check me on this.

This generates the following output:
image

Additionally may want to output a color bar to be more clear on color meanings, unless we are intending to assume the scikit-learn example output is well understood.

Purg (commented):
Similar to the change note comment, I would suggest being a little more descriptive in the name of the notebook file/title.

examples/Scikit.ipynb
Purg: Adding a follow-on MLP take sounds like a good idea to me. In adding the MLP example this into *this* notebook, I don't think we need to reproduce the reference example, just refer to it (since their results are there). We can simply create the `MLPClassifier` instance, fit it, and apply the model as a blackbox to our "app" for saliency visualization.

Side note, even their example states that the MLP model will warn that it does not converge, so it doesn't surprise me that their visualization might not be that great (it's also just the first layer out of 50, so I'm not sure what they were expecting to get out of that visualization...). We can either duplicate their hyperparameters to provide a comparable result, or tweak the iteration/learning-rate to "better" the model for our visualization. I might suggest the former for comparability while noting that we understand that non-convergence is happening.

examples/Scikit.ipynb
schencej: I think I fixed this? No totally sure.

schencej:
I swapped out the logistic regression for the MLP and I think the narrative is much better. However, there is a lot more negative saliency in each map with this approach and the averaging makes a lot of the positive values hard to see when using the [-1,1] scale. Currently I only plot the positive saliency and I scale each individual map to span from [0, 1] in order to increase visibility. Let me know if you think this is appropriate.

examples/MNIST_scikit_saliency.ipynb
```diff
@@ -0,0 +1,392 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "id": "5136a845",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "class MNISTClassifier (ClassifyImage):\n",
+    "    def get_labels(self):\n",
+    "        return [i for i in range(10)]\n",
```
Purg, line 269:
```suggestion
    "        return list(range(10))\n",
```

examples/Scikit.ipynb
Purg: Partially: it's now set to return an iterator, however the parent API states an iterator of dictionaries. Also, images are documented to be input as 2- or 3-dim. A more "correct" implementation might be:
```python
class MNISTClassifier (ClassifyImage):
    ...

    def classify_images(self, image_iter):
        # Yes, "images" in this example case are really 1-dim (28*28=784).
        # MLP input needs a (n_samples, n_features) matrix input.
        images = np.asarray(list(image_iter))  # may fail because input is not consistent in shape
        images = images.reshape(-1, 28*28)  # may fail because input was not the correct shape
        return (
            {idx: p_i for idx, p_i in zip(range(10), p)}
            for p in mlp.predict_proba(images)
        )
```

Purg:
> ... However, there is a lot more negative saliency in each map with this approach and the averaging makes a lot of the positive values hard to see when using the [-1,1] scale. Currently I only plot the positive saliency and I scale each individual map to span from [0, 1] in order to increase visibility. Let me know if you think this is appropriate.

Saliency is not just about what "positively" contributes to a prediction: knowing about what a model down-weights can matter, too. I'll leave this more for @brianhhu to validate, but maybe in this case we are seeing that the trained MLP classifier here has learned less about what parts of the input "contributes" to a class prediction, but what *does not* contribute. E.g. it might be more predicting an input is not "zero" because there are active pixels in the center, not that there isn't input in a "zero" shape (if that makes sense written down...).

If @brianhhu also thinks this is a correct interpretation maybe we can use this as a teachable moment about the positive/negative saliency dichotomy and extend the narrative of this example.

schencej:
I added back in the regression content and reorganized a bit.

I also added a blurb in the "Calling the Application" section in the "Regression" part explaining the connection between the saliency map and the coefficients.

Please take a look.

brianhhu (commented):
LGTM, the narrative is much improved! I made a few suggested edits for clarification and readability.

examples/MNIST_scikit_saliency.ipynb
```diff
@@ -0,0 +1,544 @@
...
...i8j/jD9/JD7+moh8Ku5gf11EvllErIj8aPz7/4jIJ0kcxUWN/18C/kUI4Ul8zwcv8RwvM5fh/v9u4MfP/cwuPxc19gHYjr/vAL/3ks7vsnNR4/81wC/E338R+DMv6fyOJXl+Eh8HPn2C5z0A/mQIYRG9Bj8O/EHgzwM/F0L4ByJigRHw9cDrIYSPA4jIrSOO+WdF5FuA3wE+GUJ42Z6my8hFjf9H4v/9d8ACPxBC+NkXPJeryEXe/4jIm8AHWS0GN4mLGvsfAH4+etzGwJ94wfO4qlzU+P8a8J3APwW+A9gSkbshhEcveD6nIhk/iZOSAf9cRL4eaIiLJ/A/gB8RkQz46RDCr4rI54APicg/A/4r8POHHO+/AD8eQliKyF8Bfgz4tpd9EleY8x5/B3wY+FbgDeBTIvK1IYTdl3oWV5fzHv+W7wL+YwiheXkf/cpz3mP/3cCPhhD+kYh8E/DvROTjIWy2nE9Eznv8/0483ieATwFfisd9T0lhr8RvAH/gBM/7JPAV4OtQqz8HCCF8CvgW9Ab+URH53hhK+Trgl4C/Cvzw5sFCCI9CCMv45w+f8DNcRy5k/IG3gf8cQqhCCP8P9b59+MVO5UpyUePf8l3czJAXXNzY/wXgJ+MxfgUogHsvciJXlIua+38vhPCdIYRvAP5efGz3RU/mtCTjJ/ELwEBE/nL7gIj8fhH55o3n7QBfjruj70FDJa3b/ishhB9Cb/RvFJF7gAkh/BSaXPiNm28qIq/1/vx24DfP8ZyuEhcy/sBPo14f4vM/AnzuHM/rqnBR44+IfBS4DfzKOZ/TVeGixv4LwB+Px/gYavw8PNczuxpc1Nx/T0Ra2+P7gR855/M6ESnsdcMJIQQR+Q7gn4jI3wUWwOeBv7Xx1H8J/JSIfC/ws8A0Pv6twPeJSAUcAN8LvA78m40bfJO/KSLfDtTAY+AT53RKV4oLHP+fA/6UiHwGdTl/33sdc78MXOD4g3p9fiLcUJn9Cxz7vw38UEzGDcAnbuI1uMDx/1bgB0UkoGGvv3ZOp3QqUnuLRCKRSCQSN4oU9kokEolEInGjSMZPIpFIJBKJG0UyfhKJRCKRSNwokvGTSCQSiUTiRpGMn0QikUgkEjeKZPwkEolEIpG4USTjJ5FIJBKJxI0iGT+JRCKRSCRuFP8fSAvsxam7CvoAAAAASUVORK5CYII=\n",
```
brianhhu, line 340: Can you check these output saliency maps? Somehow the saliency maps for all digits seemed to have collapsed into a single output, which is different from your previous commits. I couldn't find any obvious changes with the `app` function itself, so it may just need an output refresh.

examples/MNIST_scikit_saliency.ipynb
```diff
@@ -0,0 +1,544 @@
...
...Puuc+4yZvUjS30i6uIX7m5JOafn5dUmDzrlPjMxwzt3lnPtOfqGQzX7HzO4Ir9PC/IPM7NbwX+LdZnaGmVXN7Jow/TMzu1yYDu2K5e9I+phzbmvY55MtPMeUzITv5mskXdf0M0tPu2LpJM0L5fmSNrTo/FLTrngeI+lboXyLpFe06PwOSEe0/Eg6TtJPJrDck5Je4pzbG/7Tv07ScyRdJOlrzrn3mVlVUr+kkyQd4pw7TpLMbMEY23yVmb1A0n2SLnfOtbqlqezaFcujQ91tkqqSrnDO3XyA54L2fjdlZodLOkLZL1tMXbtieYWkr4cWvNmSzjrA84DXrnjeJemVkj4s6XxJc81ssXNuywGeT1N1SvIzUd2SPmpmJ0mqKfzBk/QjSZ8ys25JNzrn7jSzhyT9ipl9RNJNkr7eYHv/Kek659yAmf2upM9IelGrTwKSmh/LLklHSTpT0qGSbjWz451z21p6FhjR7HiOuFDS9c65WusOHQXNjuVrJF3jnPuAmT1P0ufM7Djn3HDLzwRS8+P5R2F7qyXdKumxsN0ZpVO6vX4u6dcmsNzlkjZKOlE+c+2RJOfcrZJeIB+Ea8zsktD9caKkNZLeJOnq4sacc1uccwNh8uoJHgPG15ZYSvqlpC875wadcw/Lt+QddWCnArUvniMuFF1ezdKuWL5B0hfDNm6X1CdpyYGcCCS17+/mBufcK51zJ0t6d5i37UBPptk6Jfn5lqReM3vjyAwzO8HMzigsN1/S4+E/hovluzdGmsY3Oueukg/Ws81siaSKc+4G+Qvunl3cqZkdlJs8V9I9TTynVLUllpJulG/1UVj+aEkPNfG8UtWueMrMflXSQkm3N/mcUtWuWD4i6cVhG8+ST342NfXM0tSuv5tLzGwkt3iXpE81+byaoiO6vZxzzszOl/QhM3unpL2S1kl6W2HRj0u6wcwukXSzpF1h/pmS3mFmg5J2SrpE0iGSPl0IUtEfmNm5koYkPSVpdZNOKVltjOXXJJ1tZr+Qb4J9x0zrg+5EbYyn5Ft9vuC4TX1TtDGWb5d0Vbh41klaTUwPXBvjeaakvzEzJ9/t9eYmnVJT8XgLAACQlE7p9gIAAGgKkh8AAJAUkh8AAJAUkh8AAJAUkh8AAJAUkh8AAJAUkh8AAJAUkh8AAJCU/w8CTjAOC5OQZAAAAABJRU5ErkJggg==\n",
```
brianhhu, line 497: Similarly, please check the output saliency maps here too. Thanks!

.gitignore
```diff
@@ -1,3 +1,6 @@
+# Example Data
+examples/scikit-learn-example/
```
brianhhu, line 2: Can you add this entry to the [.gitignore file within the examples directory](https://github.com/XAITK/xaitk-saliency/blob/master/examples/.gitignore) rather than the overall repo? i.e. add a new line like:

```
/scikit-learn-example/**
```

brianhhu (commented):
Please take a look at the minor review comments. It would also be nice to add Colab support by adding a simple header to the top of the notebook. Feel free to take a look at the other examples for how to do so. Thanks!

examples/MNIST_scikit_saliency.ipynb
schencej: Not sure how I missed that... it's fixed now

schencej: Not sure how I missed that... it's fixed now

examples/.gitignore
```diff
@@ -2,3 +2,4 @@
 /viame-example/**
 **/imagenet_classes.txt*
 /*.jpg
+/scikit-learn-exampel/**
```
brianhhu, line 5: `example` is mispelled, and can you move this up to where the other example directories are defined?

```suggestion
/scikit-learn-example/**
```

brianhhu (approved):
LGTM, I think technically only `xaitk-saliency` needs to be installed in the first cell (since it will pull in the other dependencies), but that's a minor issue.

examples/MNIST_scikit_saliency.ipynb
```diff
@@ -0,0 +1,572 @@
...
+     "text": [
+      "Traceback (most recent call last):\n",
+      "  File \"<string>\", line 1, in <module>\n",
+      "ModuleNotFoundError: No module named 'xaitk_saliency'\n",
+      "\u001b[33mWARNING: Discarding https://files.pythonhosted.org/packages/df/86/3db1d064d8a04fae4d4e23262a1894e27de31a490e60d21324452bac7f14/xaitk_saliency-0.3.1-1.tar.gz#sha256=a8cee265686db1422f1a0cb77b68a97c5895d168aaaa62cb14e7d21ebfcc592e (from https://pypi.org/simple/xaitk-saliency/) (requires-python:>=3.7,<3.10). Requested xaitk-saliency from https://files.pythonhosted.org/packages/df/86/3db1d064d8a04fae4d4e23262a1894e27de31a490e60d21324452bac7f14/xaitk_saliency-0.3.1-1.tar.gz#sha256=a8cee265686db1422f1a0cb77b68a97c5895d168aaaa62cb14e7d21ebfcc592e has inconsistent version: filename has '0.3.1.post1', but metadata has '0.3.1'\u001b[0m\n"
+     ]
+    }
+   ],
+   "source": [
+    "!python -c \"import xaitk_saliency\" || pip install -q \"xaitk-saliency\""
```
Purg, line 58: This error maybe suppressed by adding a STDERR redirect in the lhs command:
```python
!python -c "import xaitk_saliency" 2>/dev/null || pip install -q "xaitk-saliency"
```

Of course, checking if it is installed is a little overkill as a simpler invocation of `!pip install -q xaitk-saliency` will be a no-op when the package is already installed.

examples/MNIST_scikit_saliency.ipynb
```diff
@@ -0,0 +1,572 @@
...
+    "from sklearn.model_selection import train_test_split\n",
+    "from sklearn.linear_model import LogisticRegression\n",
+    "\n",
+    "t0 = time.time()\n",
+    "\n",
+    "# Split data into test and train sets\n",
+    "train_samples = 20000\n",
+    "\n",
+    "X_train, X_test, y_train, y_test = train_test_split(\n",
+    "    X, y, train_size=train_samples, test_size=10000)\n",
```
Purg, line 246: I just realized that this could use a fixes random seed to result in consistent results from run to run
```suggestion
    "X_train, X_test, y_train, y_test = train_test_split(\n",
    "    X, y, train_size=train_samples, test_size=10000, random_state=0)\n",
```

examples/MNIST_scikit_saliency.ipynb
```diff
@@ -0,0 +1,572 @@
...
+    "\n",
+    "# Define model, lower C value gives higher regulation\n",
+    "clf = LogisticRegression(\n",
+    "    C=50. / train_samples, penalty='l2', solver='saga', tol=0.1\n",
+    ")\n",
```
Purg, line 251: This could also use a random state seed
```suggestion
    "clf = LogisticRegression(\n",
    "    C=50. / train_samples, penalty='l2', solver='saga', tol=0.1, random_state=0\n",
    ")\n",
```

Purg (approved):
LGTM. The following comments are minor and should not block acceptance.

examples/MNIST_scikit_saliency.ipynb
brianhhu: Do you know what is the cause of the output message? In my local poetry dev environment with jupyter notebook installed, running this cell is effectively a noop.

Purg: The warning is the failure of the `python -c "import xaitk_saliency"` call, which is expected if the package is not installed. As mentioned above, we don't *need* this check since the pip install is idempotent.

Purg:
Thanks for the updates! LGTM and merging

PR #60, opened 2021-09-20
Dev/roddy docs whitebox

collinsr:
Draft text for introduction of white box methods ready for review.


docs/introduction.rst
```diff
@@ -82,27 +82,66 @@ approaches to this:
    output for the related images, we can deduce certain aspects of the
    model's behavior.
 
+
+.. note:: These descriptions apply only to *using the model after it has been created*; nothing is implied about how the model is constructed.
+
+.. note:: This distinction applies to both the *concept* and *implementation*. A model could support white box reasoning but lack the APIs to access the internal state; conversely, it might provide complete API access yet provide no guidance on how that state influences the output. **API access to the model does not automatically imply white box XAI.**
```
Purg, line 88: Minor and maybe subjective: the use of "XAI" and the end here didn't make sense / flow for me. Maybe instead the following?
```suggestion
.. note:: This distinction applies to both the *concept* and *implementation*. A model could support white box reasoning but lack the APIs to access the internal state; conversely, it might provide complete API access yet provide no guidance on how that state influences the output. **API access to the model does not automatically imply white box explainability.**
```

Purg, line 88: Can notes take multi-line strings? I might suggest a once-sentence-per-line if possible for future diff sanity.

docs/introduction.rst
```diff
@@ -82,27 +82,66 @@ approaches to this:
...
-*Explanation options are correlated to how the model operation can be
-inspected and interpreted. pros: closely tied to the individual model, thus can leverage
-specific knowledge about its architecture. Measures the actual
-computation which generates the output. Cons: specific to individual
-models / classes of models; harder to generalize across models.
-Requires modifying the model implementation to gain access; may
-require updating as the implementation evolves.*
+A **white box** model exposes some (or all) of the internal state of
+the model; the explanation draws a connection between this exposed
+state and the model output. Some models are intrinsically introspective to the point where they are not so much "white box" as transparent:
```
Purg, line 97: I think white-box is more descriptive of the explanation method than model transparency: a white-box method requires/implies some model transparency, but some model transparency does not imply it is sufficient for a white-box method.
```suggestion
A **white box** explanation method makes use of a model's exposed internal state or representation;
the explanation draws a connection between this exposed
state and the model output. Some models are intrinsically introspective to the point where they are not so much "white box" as transparent:
```

Purg (approved):
Just some minor comments that I don't think should block this if it comes down to that.

Purg:
Missing a release note though.

docs/introduction.rst
```diff
@@ -33,7 +33,7 @@ Image Saliency Maps: An Intuitive Introduction
 ----------------------------------------------
 
 Figure 4 shows a deep learning pipeline for recognizing objects in
-images; pixels in the image (in green) are processed by the pipeline
+images; pixels in the image (in green) are processed by the pipeline (in yellow)
```
melster1010, line 36: do we care about what the other colors (turquoise & red) represent in the diagram?

docs/introduction.rst
```diff
@@ -82,27 +82,66 @@ approaches to this:
...
-specific knowledge about its architecture. Measures the actual
-computation which generates the output. Cons: specific to individual
-models / classes of models; harder to generalize across models.
-Requires modifying the model implementation to gain access; may
-require updating as the implementation evolves.*
+A **white box** model exposes some (or all) of the internal state of
+the model; the explanation draws a connection between this exposed
+state and the model output. Some models are intrinsically introspective to the point where they are not so much "white box" as transparent:
+
+* In `linear regression
```
melster1010, line 99: insert a comma after "regression"

docs/introduction.rst
```diff
@@ -82,27 +82,66 @@ approaches to this:
...
+  <https://en.wikipedia.org/wiki/Linear_regression>`_ the influence of
+  the observations on the solution are an immediate consequence of the
+  method.
+
```
melster1010, line 103: s/b the "influence of the observations on the solution **is** an immediate consequence" (unless you mean the observations themselves are immediate *consequences*)

docs/introduction.rst
```diff
@@ -82,27 +82,66 @@ approaches to this:
...
+  the output is directly computed by making the comparisons and
+  branches encoded in the model.
+
+An example of a white box explanation method for CNNs is Grad-CAM (`paper <https://arxiv.org/abs/1610.02391>`_, `code <https://github.com/ramprs/grad-cam/>`_), which exposes some (but not all) of the CNN layers to the explanation generation algorithm.
+
+.. figure:: figures/intro-grad-cam-annotated.png
+
+   Figure 5: Grad-CAM architecture. Functional regions annotated to align with Figure 4: input in green, CNN architecture in yellow, output in orange. The explanation functionality is in blue; the APIs granting access to the model are in red. Figure from the `Grad-CAM code repo. <https://github.com/ramprs/grad-cam/>`_
+
+In Grad-CAM, first the model computes its output per the typical processing flow: input (green), through the CNN (yellow), to the output (orange.) The explanation is created via an additional processing step which uses the output and feature maps from within the CNN (red circles) to measure and visualize the activation of those regions associated with the output (blue).
```
melster1010, line 114: processing step *that* uses the output

melster1010: "to measure and visualize the activation of those regions associated with the output (_blue_)" 
> I thought the output was orange?

melster1010 (commented):
As a non-expert, I found this explanation clear and easy to follow. Maybe after the black blox section is done, we could also include a high-level matrix of the differences/pros & cons for easy comparison?

docs/introduction.rst
brianhhu: I think what we are trying to capture is the difference between "white box" and "black box" testing (but essentially from a ML versus software perspective)? e.g. see comparison tables [here](https://www.geeksforgeeks.org/differences-between-black-box-testing-vs-white-box-testing/) and [here](https://www.guru99.com/back-box-vs-white-box-testing.html)

docs/introduction.rst
```diff
@@ -82,27 +82,66 @@ approaches to this:
...
+
+* **The explanation could not have been created from the output alone**. In order to operate, the explanation algorithm (blue) required access to both the output *and* the CNN internals.
+* **The algorithm implementation is tightly coupled to the CNN architecture exposed by the API**. Although the *method* may be general, any particular *implementation* will require the CNN architecture to conform to the specifics of the API.
+
+In general, pros and cons of white box approaches are:
+
+Pros
+""""
+
+* A white box model can choose to **leverage its tight coupling to the model** to maximize the information available, at the sacrifice of generalization to other models.
```
brianhhu, line 126: model -> explanation?

docs/introduction.rst
```diff
@@ -82,27 +82,66 @@ approaches to this:
...
+* A white box model **accesses the actual computation which generated the output**. The explanation is derived directly from what the model computed about the input, in contrast to black box models which can only indirectly compare the output to output from slightly different inputs.
```
brianhhu, line 127: model -> explanation?

docs/introduction.rst
```diff
@@ -82,27 +82,66 @@ approaches to this:
...
+In Grad-CAM, first the model computes its output per the typical processing flow: input (green), through the CNN (yellow), to the output (orange.) The explanation is created via an additional processing step which uses the output and feature maps from within the CNN (red circles) to measure and visualize the activation of those regions associated with the output (blue).
+
+Two aspects typical of white box methods are demonstrated here:
+
+* **The explanation could not have been created from the output alone**. In order to operate, the explanation algorithm (blue) required access to both the output *and* the CNN internals.
+* **The algorithm implementation is tightly coupled to the CNN architecture exposed by the API**. Although the *method* may be general, any particular *implementation* will require the CNN architecture to conform to the specifics of the API.
+
+In general, pros and cons of white box approaches are:
+
+Pros
```
brianhhu, line 123: Another pro is generally only requires a single forward/backward pass through the model, so it is more computationally efficient compared to black-box explanation methods.

brianhhu (commented):
Overall, I think this is great progress. The main issue I think we need to sort out is the semantic distinction between white box models and explanations. We seem to be using these somewhat interchangeably, which introduces some confusion.

docs/introduction.rst
collinsr: fill-paragraph-or-region seems to indent correctly for multi-line notes; applied to both.

docs/introduction.rst
```diff
@@ -33,7 +33,7 @@ Image Saliency Maps: An Intuitive Introduction
 ----------------------------------------------
 
 Figure 4 shows a deep learning pipeline for recognizing objects in
```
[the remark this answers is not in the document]
collinsr: Yes, good point... done.

docs/introduction.rst
collinsr: No; I replaced the underlying image with an uncolored version for clarity.

docs/introduction.rst
collinsr: done

docs/introduction.rst
collinsr: done

PR #64, opened 2021-09-30
Add example notebook for saliency on Atari deep RL agent

schencej:
Here is the first draft of the deep RL example. The resulting saliency maps are not as clean as what the authors of the paper got. The notable differences that I can find between our implementations are the following:
- Their perturbation method blurs a radial section of the frame, ours fills with a single color
- They perturb the pre-processed images of size (80,80) and resize the resulting saliency maps back to full frame size where as I am perturbing the original frame and then pre-processing for model input

I don't think we have the functionality to blur as they do but I will try performing the perturbation on the pre-processed frames and see what that changes.
Please take a look and let me know what you think


schencej:
I changed the notebook to perturb the smaller, processed images and to use blurring and the results look much better I think. Please take a look.

PR #66, opened 2021-10-01
Added a workflow for CI test running example notebooks

Purg:


examples/VIAME_OcclusionSaliency.ipynb
```diff
@@ -47,10 +47,14 @@
...
-    "!python -c \"import torch\" || pip install \"torch==1.9.0\"\n",
-    "!python -c \"import torchvision\" || pip install \"torchvision==0.10.0\""
+    "!pip install -U pip\n",
+    "!pip install xaitk-saliency\n",
+    "!pip install \"torch==1.9.0\"\n",
+    "!pip install \"torchvision==0.10.0\"\n",
+    "!pip install git+https://gitlab.kitware.com/viame/bioharn.git@dev/0.2.0  # test with 0.2.0\n",
+    "# Remove opencv-python, which required libGL, which we don't require here, and replace with opencv-python-headless\n",
+    "!pip uninstall -y opencv-python opencv-python-headless  # make sure they're both gone.\n",
+    "!pip install opencv-python-headless"
```
brianhhu, line 57: In the VIAME and scikit-learn notebooks, we simplified the first cell to essentially use a `pip install` to check for packages, which is a no-op if the package is already present. Should we change this for the other notebooks for consistency (they currently use a more complex `python -c import` pattern)?

e.g. DRISE, OcclusionSaliency, SimilarityScoring, SuperPixelSaliency, covid_classification

examples/README.md
```diff
@@ -42,6 +42,18 @@ Some notebooks may require additional data. This data will be downloaded when ru
...
+### Continuous Integration
+Some notebooks contained here are executed as part of the continuous
+integration workflows to ensure their example appropriately reflects the
+toolkit's current functionality.
+Not all are included in this CI due to their long-running or heavy resource
+requirements, e.g. the ``covid_classification.ipynb`` which trains a deep
+learning model which at the very least takes a non-trivial amount of time.
+
+The CI workflow may be found in ``.github/workflows/ci-example-notebooks.yml``.
+The specification of which notebooks are included in CI is around L29.
```
brianhhu, line 55: The documentation here is great- do you think we need to update `docs/review_process.rst` in our ReadTheDocs to account for notebooks also being covered by CI? Or stated another way, what should a contributor who adds a demo notebook expect in terms of the review process?

.github/workflows/ci-example-notebooks.yml
```diff
@@ -0,0 +1,80 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        notebook-file: [
+            "DRISE.ipynb",
+            "MNIST_scikit_saliency.ipynb",
+            "OcclusionSaliency.ipynb",
+            "SimilarityScoring.ipynb",
+            "SuperPixelSaliency.ipynb",
+            "VIAME_OcclusionSaliency.ipynb",
```
brianhhu, line 35: Just to check my understanding- is this workflow run on every PR? Do you anticipate that we will hit any data usage limits if the number of covered notebooks grows and we have to run through every notebook?

brianhhu (commented):
LGTM, added a few minor comments.

examples/VIAME_OcclusionSaliency.ipynb
Purg: We probably should, but I initially considered that out of scope for this PR. If we want to just "get it done" I can make that update, but it's just lower on the local priority list.

examples/README.md
Purg: Valid point. I can add an info section to the review-process doc, including a note for the human reviewer to check for CI inclusion of new notebooks. It sounds like we should add to the ``CONTRIBUTING.md`` about adding notebooks to the CI upon that sort of contribution.

.github/workflows/ci-example-notebooks.yml
Purg: > Just to check my understanding- is this workflow run on every PR?

Yes, as well as every modification of the PR.

> Do you anticipate that we will hit any data usage limits if the number of covered notebooks grows and we have to run through every notebook?

I don't currently know what aspects of our notebooks will be applicable to a data-cap. Some aspects like downloading an image from flikr or torch are relatively small and infrequent. OpenML dataset might be an exception. If we hit slowdowns or data-limits we can enable more aggressive caching on the action side to drastically bring down the frequency of download requests/requirements.

I also set up this list to be explicit for now to give us an axis of control if such a described issue comes to the forefront to not block *all* notebook CI if one or two become problem children.

I'm not opposed to an alternative recommendation if someone comes in with one.

examples/VIAME_OcclusionSaliency.ipynb
brianhhu: Sounds good, I'll add it to the board as something to tackle in another PR.

.github/workflows/ci-example-notebooks.yml
brianhhu: Sounds good, I don't have any better ideas, so we can go with this for now. One thing to note is that the VIAME notebook downloads a 5 GB model file (from viame.kitware.com), which is something that we might want to revisit.

CONTRIBUTING.md
```diff
@@ -37,12 +37,49 @@ See the [XAITK-Saliency README](README.md) file for additional information.
...
+When contributing a Jupyter notebook, as an example or otherwise, a reference
+should be added to this CI workflow ([located here ~L27](
+.github/workflows/ci-example-notebooks.yml)) to enable its inclusion in the CI
+testing.
+
+To that end, in developing the notebook, consider its execution in this CI
+environment:
+* should be concise in its execution time to not stall or time-out the CI
+  workflow.
+* should be light on computational resources to not exceed.
```
brianhhu, line 81: Seems like this bullet point is missing a word to complete the thought?

docs/review_process.rst
```diff
@@ -132,6 +132,13 @@ by the current host ReadTheDocs.org.
 
 Passage of these checks is strictly required.
 
+Example Notebooks Execution
+---------------------------
+This check executed included example notebooks to ensure they proper
```
brianhhu, line 137: Minor edit:

```suggestion
This check executed included example notebooks to ensure their proper
```

brianhhu (approved):
LGTM, made a few minor word change suggestions.

CONTRIBUTING.md
Purg: Yup, I didn't finish that.

PR #67, opened 2021-10-01
Add SlidingRadial image perturbation implementation and docs

schencej:


schencej:
The radius of the occluded area is not preserved when you use blurring. Not sure if this is the behavior we want or not.

brianhhu (commented):
@Purg Please take a look- if we want to have a "correct" implementation of the ATARI-based saliency, I think we need to get this `SlidingRadial` perturber in first.

docs/release_notes/pending_release.rst
```diff
@@ -42,5 +45,8 @@ Examples
 * Add example notebook using classifier-based interfaces and implementations
   with scikit-learn on the MNIST dataset.
 
+* Add example notebook performing saliency on a deep RL agent in an Atari 2600
```
Purg, line 48: This is not applicable for this PR.

xaitk_saliency/impls/perturb_image/sliding_radial.py
```diff
@@ -0,0 +1,84 @@
...
+            mask = np.zeros((img_h, img_w))
+            circle = (x-center_x)*(x-center_x) + (y-center_y) * \
+                (y-center_y) <= self.radius*self.radius
+            mask[circle] = 1
+
+            if self.sigma:
+                mask = gaussian_filter(mask, sigma=self.sigma)
+                mask = mask/mask.max()
+
+            masks[i] = 1 - mask
```
Purg, line 66: In the loop, we are allocating new memory (`np.zeros((img_h, img_w))`) each time to create a layer of the output masks each time on top of the already allocated all-masks matrix. This seems wasteful as we can just work with a literal slice of the `masks` matrix instead for in-place operation. Additionally, if we just initialize `masks` to ones, we can avoid needing the `masks[i] = 1 - mask` line.
```suggestion
        num_masks = len(center_xs) * len(center_ys)
        masks = np.ones((num_masks, img_h, img_w), dtype='float32')
        center_xs_m = np.tile(center_xs, len(center_ys))
        center_ys_m = np.repeat(center_ys, len(center_xs))

        y, x = np.ogrid[:img_h, :img_w]

        for i, (center_x, center_y) in enumerate(zip(center_xs_m, center_ys_m)):
            mask = masks[i]
            circle = (x-center_x)**2 + (y-center_y)**2 <= self.radius**2
            mask[circle] = 0

            if self.sigma:
                mask[:] = gaussian_filter(mask, sigma=self.sigma)
                mask[:] = mask/mask.max()
```

This results in a "different" mask output, but that's due to seeming floating point error. The absolute summed difference between the original and this impl's output is on the order of `1e-5`.

xaitk_saliency/impls/perturb_image/sliding_radial.py
```diff
@@ -0,0 +1,84 @@
...
+        self.stride: Tuple[int, int] = (int(stride[0]), int(stride[1]))
+
+    def perturb(
+        self,
+        ref_image: np.ndarray
+    ) -> np.ndarray:
+        stride_h, stride_w = self.stride
+        img_size = ref_image.shape[:2]
+        img_h, img_w = img_size
+        center_xs = np.arange(0, img_w, stride_w)
```
Purg, line 46: Given a previous conversation, do we want this to start with the center of the circle in the upper left? Or with the circle fully within the image (e.g. `np.arange` start = `radius`).

docs/release_notes/pending_release.rst
brianhhu: Removed.

xaitk_saliency/impls/perturb_image/sliding_radial.py
```diff
@@ -0,0 +1,84 @@
...
+    values when summing the generated masks will not be even.
+
+    Related, if the stride is set to be larger than the circle diameter, the
+    resulting plane of summed values will also not be even, as there be
+    increasingly long valleys of unperturbed space between masked regions.
+
+    :param radius: The radius of the occlusion area.
+    :param sigma: The sigma value for the Gaussian filter applied to mask.
+    :param stride: The striding step for the center of the radial as a tuple
+        with format `(height_step, width_step)`.
```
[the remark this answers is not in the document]
brianhhu: The docstring for SlidingWindow should also probably be updated for clarity.

xaitk_saliency/impls/perturb_image/sliding_radial.py
```diff
@@ -0,0 +1,84 @@
...
+
+        num_masks = len(center_xs) * len(center_ys)
+        masks = np.empty((num_masks, img_h, img_w), dtype='float32')
+        center_xs_m = np.tile(center_xs, len(center_ys))
+        center_ys_m = np.repeat(center_ys, len(center_xs))
+
+        y, x = np.ogrid[:img_h, :img_w]
+
+        for i, (center_x, center_y) in enumerate(zip(center_xs_m, center_ys_m)):
+            mask = np.zeros((img_h, img_w))
```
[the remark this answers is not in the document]
brianhhu: Nope, you are correct in that this is more efficient- I have changed this now.

xaitk_saliency/impls/perturb_image/sliding_radial.py
brianhhu: I think we should revisit this and make a similar design choice for SlidingWindow (i.e. whether the center of the box/circle starts at 0,0 or fully includes the box/circle).

xaitk_saliency/impls/perturb_image/sliding_radial.py
```diff
@@ -0,0 +1,84 @@
...
+    def perturb(
+        self,
+        ref_image: np.ndarray
+    ) -> np.ndarray:
+        stride_h, stride_w = self.stride
+        img_h, img_w = ref_image.shape[:2]
+
+        # Center masks in between pixels
+        center_xs = np.arange(0.5, img_w, stride_w)
+        center_ys = np.arange(0.5, img_h, stride_h)
```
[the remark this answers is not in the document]
brianhhu: This discussion is captured in another comment below, so resolving.

xaitk_saliency/impls/perturb_image/sliding_radial.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+
+    def __init__(
+        self,
+        radius: int = 50,
+        sigma: int = 0,
+        stride: Tuple[int, int] = (20, 20),
+    ):
+        self.radius: int = int(radius)
+        self.sigma: int = int(sigma)
```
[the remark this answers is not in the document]
brianhhu: Do we want to extend both of these to be two-dimensional? This would allow for support of ellipse-like masks, as well as anisotropic blurring in x/y.

xaitk_saliency/impls/perturb_image/sliding_radial.py
brianhhu: Unfortunately, I think this approach generates different results for the blurring case (i.e. self.sigma > 0). Blurring a circular mask of 0's on a background of 1's is actually different than blurring a circular mask of 1's on a background of 0's and then inverting the final result. I think it has to to do with the final normalization.

Purg: If it needs to be the original form, then that's what it needs to be. My local naive testing, which included blurring, showed no discernible difference in the mask output (absolute sum total difference of ~1e-5), with literally no difference when applying the masks to uint8 scale space.

brianhhu:
@Purg  Please take a look, some notable updates:

1) Changed implementation to use [scikit-image's ellipse function](https://scikit-image.org/docs/dev/api/skimage.draw.html#skimage.draw.ellipse) for native support of ellipses (and circles).

2) Made the `sigma` value optional, with a default value of None (i.e. no blurring).

3) When no blurring is done, a binary mask is generated (previously this was float).

Purg:
@Purg, interesting note: it looks like LGTM picks up unused variable imports, while SonarCloud seems to have missed those.

Huh, that's surprising. Well, then there is something they are doing differently at least, with LGTM adding some legitimacy behind it taking so long to process.

PR #75, opened 2021-10-20
Notebook CI cacheing

Purg:


Purg:
> Also, since we are editing the notebooks here in this PR, it might make sense to tack on the change of XAITK-Saliency to xaitk-saliency, per our [Gitlab issue](https://gitlab.kitware.com/xaitk/meta/-/issues/132). I think only two notebooks were affected (mainly the scikit-learn one).

Oops, missed this. Let me fix those and then I can use that update to test any changes in notebook CI runtimes.

PR #78, opened 2021-11-03
Update top-level README with more content

Purg:


README.md
```diff
@@ -1,14 +1,61 @@
...
+<hr/>
+
+[![Documentation Status](https://readthedocs.org/projects/xaitk-saliency/badge/?version=latest)](https://xaitk-saliency.readthedocs.io/en/latest/?badge=latest)
+[![badge-unittests](https://github.com/xaitk/xaitk-saliency/actions/workflows/ci-unittests.yml/badge.svg)](https://github.com/XAITK/xaitk-saliency/actions/workflows/ci-unittests.yml)
+[![badge-notebooks](https://github.com/xaitk/xaitk-saliency/actions/workflows/ci-example-notebooks.yml/badge.svg)](https://github.com/XAITK/xaitk-saliency/actions/workflows/ci-example-notebooks.yml)
+[![codecov](https://codecov.io/gh/XAITK/xaitk-saliency/branch/master/graph/badge.svg?token=VHRNXYCNCG)](https://codecov.io/gh/XAITK/xaitk-saliency)
+[![Quality Gate Status](https://sonarcloud.io/api/project_badges/measure?project=XAITK_xaitk-saliency&metric=alert_status)](https://sonarcloud.io/summary/new_code?id=XAITK_xaitk-saliency)
+[![Language grade: Python](https://img.shields.io/lgtm/grade/python/g/XAITK/xaitk-saliency.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/XAITK/xaitk-saliency/context:python)
+
+The `xaitk-saliency` package is an open-source, explainable AI framework for
```
brianhhu, line 12: Define XAI here since we use the acronym below. We might want to also update our Github repo description to match (not strictly necessary).

```suggestion
The `xaitk-saliency` package is an open-source, explainable AI (XAI) framework for
```

README.md
```diff
@@ -1,14 +1,61 @@
...
+
+This framework is a part of the [Explainable AI Toolkit (XAITK)](
+https://xaitk.org).
+
+## Target Audience
+This toolkit is intended to help data scientists and developers who want to
+add visual saliency explanations to their workflow or product for relevant
+models.
+Functionality provided here is both directly accessible for targeted
+experimentation, and through Strategy and Adapter patterns to allow for
```
brianhhu, line 28: I'm not super familiar with "Strategy and Adapter patterns"- seems like a pretty niche term although one that was brought up by the JAIC. Should we provide additional links that define these terms?

README.md
```diff
@@ -1,14 +1,61 @@
...
+## Installation
+Install the latest release via pip:
+```bash
+pip install xaitk-saliency
+```
+
+Some plugins may require additional dependencies in order to be utilized at
+runtime.
+Such details are described [here](
+https://xaitk-saliency.readthedocs.io/en/latest/implementations.html).
```
brianhhu, line 40: Should this link to the interfaces page instead? I don't see anything about additional dependencies on the implementations page, although I could be missing something.

```suggestion
Such details are described [here](
https://xaitk-saliency.readthedocs.io/en/latest/interfaces.html).
```

README.md
```diff
@@ -1,14 +1,61 @@
...
+
+See [here for more installation documentation](
+https://xaitk-saliency.readthedocs.io/en/latest/installation.html).
+
+## Getting Started
+We provide a number of Jupyter-notebook based examples in the `./examples/`
+directory to show example usage of the `xaitk-saliency` package in a number of
```
brianhhu, line 47: We use "example" a lot in this sentence here, I would recommend dropping the last one:

```suggestion
directory to show usage of the `xaitk-saliency` package in a number of
```

README.md
```diff
@@ -1,14 +1,61 @@
...
+different contexts.
+
+Contributions are welcome!
+Please see the [CONTRIBUTING.md](./CONTRIBUTING.md) file for details.
 
 ## Documentation
-https://xaitk-saliency.readthedocs.io/en/latest/
+Documentation snapshots for releases as well as the latest master are hosted on
+[ReadTheDocs here](https://xaitk-saliency.readthedocs.io/en/latest/).
```
brianhhu, line 55: I think we can remove the "here":

```suggestion
[ReadTheDocs](https://xaitk-saliency.readthedocs.io/en/latest/).
```

README.md
```diff
@@ -1,14 +1,61 @@
-# XAITK - Saliency
+![xaitk-logo](./docs/figures/xaitk-wordmark.png)
```
brianhhu, line 1: It might make sense to still have a text header that includes the word "saliency" in some form or another, since we might eventually standardize the headers of all repos under the XAITK github namespace (e.g. autonomy, web-api, web-demo, etc.) to have this same logo image.

README.md
```diff
@@ -1,14 +1,61 @@
...
 
-## Intent
-Provide interfaces that convey a standard API for visual saliency
-map generation.
+<hr/>
+
+[![Documentation Status](https://readthedocs.org/projects/xaitk-saliency/badge/?version=latest)](https://xaitk-saliency.readthedocs.io/en/latest/?badge=latest)
```
brianhhu, line 5: Can we also add badges for the current pypi version (v0.4.0) and also python version (e.g. python 3.6/7/8)? Surveying other XAI packages, these seem to be included in their READMEs. I'm OK removing the "quality gate" and "code quality" badges if we need more space.

brianhhu (commented):
The only other thing I would recommend adding is some pointer to currently implemented methods in the package. This might essentially be a link to the section at the end of the introduction for different algorithms grouped by task (e.g. image classification, image retrieval, etc.), although I'm open to other ideas as well.

README.md
Purg: Sounds good. Added.

README.md
Purg: I thought I would add those here to be attractive for the more "developer" target audience. I can add wiki links, though those will be generally broad in relation to their particular application here.

README.md
```diff
@@ -1,14 +1,61 @@
...
+See [here](https://xaitk-saliency.readthedocs.io/en/latest/introduction.html)
+for a more formal introduction to the topic of XAI and visual saliency
+explanations.
+
+This framework is a part of the [Explainable AI Toolkit (XAITK)](
+https://xaitk.org).
+
+## Target Audience
+This toolkit is intended to help data scientists and developers who want to
+add visual saliency explanations to their workflow or product for relevant
```
melster1010, line 25: "relevant" is rather vague, can this be more specific?

README.md
Purg: I specifically meant to refer to the implementations page as they would be the level in which additional, optional dependencies would be added. Interfaces by the nature of their role should not introduce "optional" dependencies. You are correct that none do currently, but I was more mentioning this generically. If you think we should drop this for now to avoid confusion, I'm fine with that.

README.md
melster1010: "open source" is not hyphenated per the [Kitware Style Guide](https://docs.google.com/document/d/1YasFqNoaQrjC_xijluTpg6_4cgu3dWAp8x02tAdLIG0/edit) (not the xaitk style sheet, since this guideline applies to Kitware text in general)

README.md
```diff
@@ -1,14 +1,61 @@
...
+See [here for more installation documentation](
+https://xaitk-saliency.readthedocs.io/en/latest/installation.html).
+
+## Getting Started
+We provide a number of Jupyter-notebook based examples in the `./examples/`
+directory to show example usage of the `xaitk-saliency` package in a number of
+different contexts.
+
+Contributions are welcome!
+Please see the [CONTRIBUTING.md](./CONTRIBUTING.md) file for details.
```
melster1010, line 51: Omit the "Please" (I think we're friendly enough with the preceding exclamation mark :)

README.md
Purg: I've removed the "Quality Gate Status" and added pypi latest-release version and supported python versions badges. Much more than this seems likely superfluous for now.

melster1010 (commented):
A few comments, mostly style stuff.

README.md
brianhhu: Sounds good to me- thanks!

README.md
Purg: I don't have anything more specific so I'll just remove the tail of that sentence.

README.md
```diff
@@ -1,14 +1,61 @@
...
+Some plugins may require additional dependencies in order to be utilized at
+runtime.
+Such details are described [here](
+https://xaitk-saliency.readthedocs.io/en/latest/implementations.html).
+
+See [here for more installation documentation](
+https://xaitk-saliency.readthedocs.io/en/latest/installation.html).
+
+## Getting Started
+We provide a number of Jupyter-notebook based examples in the `./examples/`
```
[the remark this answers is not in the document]
Purg: :+1:

README.md
Purg: :+1:

Purg:
Yes, I realize I'm pushing fix-up commits. Their being separate commits just helps with seeing differences from review to review. I'm waiting until an approving review before I autosquash them.

README.md
brianhhu: Thanks for the clarification, let's keep this as is for now.

brianhhu (approved):
LGTM, thanks!

PR #79, opened 2021-11-04
Normalize saliency maps

schencej:
The saliency maps are now normalized inside the generator function. The maps are then resized and scaled by 255 so that a value of 1 will give a blue/red value of 255 in the visual saliency representation. This operation replaces the post-processing function.


examples/atari_deepRL_saliency.ipynb
```diff
@@ -206,7 +206,7 @@
      "name": "stdout",
```
brianhhu, line 206: Since we are editing this notebook, can we remove the "please" in the header section when talking about Colab? I think this was one of the changes that Mary Elise had wanted to make.

"To run this notebook in Colab, please use the link below:" -> "To run this notebook in Colab, use the link below:"

brianhhu (commented):
Left a few minor comments, mainly about reconsidering the use of post-processing function. @Purg, can you also take a look? This PR does not add a new interface/implementation to the base package, but only updates the notebook.

examples/atari_deepRL_saliency.ipynb
```diff
@@ -457,21 +457,6 @@
...
-    "    Helper function for post-processing saliency maps, which includes resizing\n",
-    "    the saliency maps to (160,160) and multiplying them by a \"fudge factor\" ff\n",
-    "    as in the original implementation.\n",
-    "    \"\"\"\n",
-    "    pmax = sal.max()\n",
-    "    sal = cv2.resize(src=sal, dsize=(160,160)).astype(np.float32)\n",
-    "    sal -= sal.min()\n",
-    "    sal = ff * pmax * sal / sal.max()\n",
-    "    \n",
-    "    return sal\n",
```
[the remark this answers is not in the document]
Purg: I thought we were trying to get rid of fudge-factors and the like with this revision? If you think it best to keep those in here for the sake of visualization, then OK, but we should be clear that this function is for that only. Ideally we would have a good explanation, or a link to one, for the effect of the fudge factor on the visualization, otherwise I don't know if it's just creating an illusions of success.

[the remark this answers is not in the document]
brianhhu: I don't think it's possible to remove the fudge factor and still be faithful to the original implementation. However, I do believe that what Elim has now looks OK (without adding a fudge factor), so I'm happy to go with that if we don't think it's critical to match the original implementation exactly.

examples/atari_deepRL_saliency.ipynb
```diff
@@ -457,21 +457,6 @@
...
-    "\n",
-    "\n",
     "class PolicyScoring(GenerateClassifierConfidenceSaliency):\n",
     "    \"\"\"\n",
     "    This saliency implementation transforms black-box action and value scores\n",
```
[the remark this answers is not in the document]
brianhhu: We need a new implementation here since the saliency algorithm (proposed by Greydanus et al., '18) takes a squared difference between reference and perturbed scores, which is something that doesn't currently exist in our other classifier implementations.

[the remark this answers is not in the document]
schencej: To quote the authors:

> We are interested in answering the question, “How much
does removing information from the region around location
(i, j) change the policy?"

They wanted to visualize the overall effect without discerning between positive and negative. They chose squared different for this.

schencej:
I've added the `PolicyScoring` implementation into the toolkit and changed the example to use that instead. As we discussed briefly in the meeting, I think an implementation that uses squared difference could apply to more than just policy networks. Perhaps we should rename and re-document.

I also changed the visualization a bit in the example. It's now handled by a separate function to emphasize that everything that is done to the saliency maps is purely for visualization. The maps are now alpha blended with the game frames instead of manipulating each frame's pixel intensities directly. This results in similar visualizations and doesn't require as much hacky stuff.

docs/implementations.rst
```diff
@@ -48,6 +48,11 @@ Class: RISEScoring
 .. autoclass:: xaitk_saliency.impls.gen_classifier_conf_sal.rise_scoring.RISEScoring
   :members:
 
+Class: PolicyScoring
```
brianhhu, line 51: I think we decided to list implementations in alphabetical order, so please adjust this order appropriately based on whatever final name for the implementation we decide upon.

xaitk_saliency/impls/gen_classifier_conf_sal/policy_scoring.py
```diff
@@ -0,0 +1,53 @@
...
+        diff = 0.5*((reference - perturbed)**2).sum(axis=1, keepdims=True)
+        sal = weight_regions_by_scalar(diff, perturbed_masks)[0]
+
+        # Normalize saliency map to [0,1]
+        sal -= sal.min()
+        sal = sal / sal.max()
+
+        return sal
+
+    def get_config(self) -> dict:
```
brianhhu, line 52: Looks like codecov complains about a missing test for the configuration portion of this new implementation- I think this should be relatively straightforward to add using a similar test from another implementation.

examples/atari_deepRL_saliency.ipynb
```diff
@@ -83,6 +83,14 @@
      "text": [
       "Action space: ['NOOP', 'FIRE', 'RIGHT', 'LEFT']\n"
      ]
+    },
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "A.L.E: Arcade Learning Environment (version +a54a328)\n",
+      "[Powered by Stella]\n"
```
brianhhu, line 92: I noticed that this new output was added- are you still using the pinned version of `open-gym` (v. 0.19.0)? I ran into some issues with installing newer versions of `open-gym`, as I believe they updated how the ATARI games were provided with the installation in more recent versions. As a result, it looks like the generated frames/videos are also different in this update compared to those in the original notebook. I purposefully chose frames/clips (taking advantage of the determinism of setting a seed) that highlighted the "tunneling" strategy of the agent.

examples/atari_deepRL_saliency.ipynb
```diff
@@ -600,15 +582,9 @@
     "    for img_idx in range(start_frame, end_frame+1):\n",
     "        sal_idx = img_idx-start_frame\n",
     "        \n",
-    "        # Post-processing here follows Greydanus et al., '18\n",
-    "        img = history['ins'][img_idx].copy().astype('uint16')\n",
-    "        \n",
-    "        # Highlight actor saliency in blue and critic saliency in red\n",
-    "        img[35:195,:,2] += postprocess_sal(actor_sal_maps[sal_idx], 300).astype('uint16')\n",
-    "        img[35:195,:,0] += postprocess_sal(critic_sal_maps[sal_idx], 600).astype('uint16')\n",
+    "        frame = visualize_saliency(history['ins'][img_idx], actor_sal_maps[sal_idx], critic_sal_maps[sal_idx])\n",
```
brianhhu, line 585: Just to confirm, do we need to create a copy of this frame history, or is it OK to pass in the original input here? We seem to create a copy when visualizing the individual frame, but not when generating the video.

e.g. using `history['ins'][img_idx].copy()`

examples/atari_deepRL_saliency.ipynb
```diff
@@ -711,14 +688,11 @@
...
+    "img = history['ins'][frame_ix].copy()\n",
     "\n",
-    "# Highlight actor saliency in blue and critic saliency in red\n",
-    "img[35:195,:,2] += postprocess_sal(actor_sal_maps[frame_ix-1020], 300).astype('uint16')\n",
-    "img[35:195,:,0] += postprocess_sal(critic_sal_maps[frame_ix-1020], 600).astype('uint16')\n",
-    "img = img.clip(1,255).astype('uint8')\n",
+    "viz = visualize_saliency(img, actor_sal_maps[frame_ix-1020], critic_sal_maps[frame_ix-1020])\n",
     "\n",
-    "plt.imshow(img)\n",
+    "plt.imshow(viz)\n",
```
brianhhu, line 695: The saliency here looks much weaker compared to what we had originally, but it could be due to the fact that those visualization tricks were used to help "saturate" the saliency and make it stronger than it actually was. It could also be due to the fact that the frame we're explaining here is different than the original frame (see the comment above about the ATARI environment and seed), which I had cherry-picked to show good saliency.

brianhhu (commented):
I added my comments in the review. Regarding naming of the new implementation, I would suggest something less RL-specific (citing the Greydanus, et al. '18 paper in the docstring still) since it is plausible that this scoring method could also be used for CV-based applications. Maybe something like `SquaredDifferenceScoring`?

Purg (commented):
I resolved a number of previous comments that look to be addressed. I don't have much else to add that hasn't already been commented on: jut a couple wording or nitpicky things that could be ignored for now or addressed in a different PR (so as to not block this one).

examples/atari_deepRL_saliency.ipynb
schencej: I believe the copy is not needed. The image is not mutated by the `visualize_saliency()` function. The `cv2.addWeighted()` function used there returns a new image.

examples/atari_deepRL_saliency.ipynb
schencej: Yeah, I was using 0.20.0. Fixed.

xaitk_saliency/impls/gen_classifier_conf_sal/policy_scoring.py
```diff
@@ -0,0 +1,53 @@
...
+    from an actor-critic policy network into saliency heatmaps. This should
+    require a sequence of either action scores or values scores predicted on
+    the reference image, a number of action or values scores predicted on
+    perturbed images, as well as the masks of the reference image perturbations
+    (as would be output from a `PerturbImage` implementation).
+
+    Although the policy network is not an image classification model per se, it
+    shares certain similarities in that it also receives as input an image
+    (i.e. a video game frame) and outputs vectors of action and/or value scores
+    which are similar to classifier confidences.
```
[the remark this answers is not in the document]
schencej: I removed this blurb after renaming the implementation.

examples/atari_deepRL_saliency.ipynb
schencej: This was indeed caused by the incorrect gym version (see above).

Alpha blending the heatmap with the game frame results in slightly less visible saliency in the visualization. I increased the alpha value of the saliency slightly to make it pop a little more.

tests/impls/gen_classifier_conf_sal/test_squared_difference_scoring.py
```diff
@@ -0,0 +1,91 @@
+from xaitk_saliency.impls.gen_classifier_conf_sal.squared_difference_scoring import SquaredDifferenceScoring
+from xaitk_saliency import GenerateClassifierConfidenceSaliency
+
+from tests import EXPECTED_MASKS_4x6
+
+import numpy as np
+import pytest
+
+
+class TestPolicyScoring:
```
Purg, line 10: This should be updated to reflect the class-rename.

schencej:
Should be good to go.

brianhhu (approved):
LGTM, thanks!

Purg (approved):
LGTM as well.

Purg:
Rebasing and updating this branch.

PR #81, opened 2021-11-23
Add function and cli tool for configurable detection sal gen

schencej:


Purg:
An initial general comment here is that the current take is focused specifically on DRISE the implementation. We should probably write to the interface, not the implementation, e.g. write the utility to use generic `GenerateDetectorProposalSaliency` implementation, the instance of which we would take in a configuration for in a similar way the `DetectImageObjects` is being used to hydrate the blackbox detector.

xaitk_saliency/utils/bin/drise_on_img_dir.py
```diff
@@ -0,0 +1,61 @@
...
+from xaitk_saliency.utils.tools.drise_from_config import drise_from_config
+
+
+@click.command()
+@click.argument('img_dir')
+@click.argument('output_dir')
+@click.argument('config_file')
+@click.option('--overlay-image/--no-image', default=False)
+@click.option('--perturbations', default=100)
+@click.option('--verbose', '-v', count=True)
```
Purg, line 17: We should have an option to output a generated config file with defaults and, more interestingly, possibly implementation sub-configs for the available plugins (right now for the `ImageObjectDetector` impls available, but also in the future the different proposal generation methods as well).

xaitk_saliency/utils/tools/drise_from_config.py
```diff
@@ -0,0 +1,130 @@
...
+
+from xaitk_saliency.impls.perturb_image.rise import RISEGrid
+from xaitk_saliency.impls.gen_detector_prop_sal.drise_scoring import DRISEScoring
+from xaitk_saliency.utils.masking import occlude_image_batch
+from xaitk_saliency.utils.detection import format_detection
+
+
+def drise_from_config(
+    ref_imgs: Collection[np.ndarray],
+    config_dict: dict,
```
Purg, line 15: This functional component should take in an actual instance of `DetectImageObjects`. The "hydration" of the instance should probably happen during the click CLI function before calling this function.

xaitk_saliency/utils/tools/drise_from_config.py
```diff
@@ -0,0 +1,130 @@
...
+        implementation.
+    :param num_perturbations: Number of perturbed images to use for the
+        saliency map generation. Defaults to 100 if not provided.
+    :param verbose: Option to print out progress statements. Default is false.
+
+    :returns: List of saliency map matrices, each with shape
+        [nDets, imgHeight, imgWidth] respective to their corresponding image.
+    """
+
+    def dets_to_formatted_mat(dets):  # type: ignore
```
Purg, line 40: We should have type annotations here.

xaitk_saliency/utils/tools/drise_from_config.py
```diff
@@ -0,0 +1,130 @@
...
+            for class_dict in [det[1] for det in dets]
+        ])
+
+        return format_detection(bbox_mat, class_mat)
+
+    # DetectImageObjects implementation
+    blackbox_detector = from_config_dict(config_dict, DetectImageObjects.get_impls())
+
+    img_perturber = RISEGrid(num_perturbations, 8, 0.5, seed=0, threads=4)
+    sal_generator = DRISEScoring()
```
Purg, line 70: In making the input `PerturbImage` and `GenerateDetectorProposalSaliency` input algorithms to this tool, the implementation choice and their configuration would be exposed for user modification.

xaitk_saliency/utils/tools/drise_from_config.py
```diff
@@ -0,0 +1,130 @@
...
+
+    # List of saliency maps for all images
+    sal_maps_set = []
+
+    for i, ref_img in enumerate(ref_imgs):
+
+        if verbose:
+            print(f'[{i+1}/{len(ref_imgs)}] Getting reference detections.')
+
+        ref_dets = iter(blackbox_detector([ref_img])).__next__()
```
Purg, line 80: The detector API is vectorized but we are not utilizing that here. Before this for-loop we should be calling the blackbox detector with the whole iterable of images and then this for-loop could take the
```
for i, (ref_img, ref_dets) in enumerate(zip(ref_imgs, detector_output_iter)):
```

xaitk_saliency/utils/tools/drise_from_config.py
```diff
@@ -0,0 +1,130 @@
...
+        return format_detection(bbox_mat, class_mat)
+
+    # DetectImageObjects implementation
+    blackbox_detector = from_config_dict(config_dict, DetectImageObjects.get_impls())
+
+    img_perturber = RISEGrid(num_perturbations, 8, 0.5, seed=0, threads=4)
+    sal_generator = DRISEScoring()
+
+    # List of saliency maps for all images
+    sal_maps_set = []
```
Purg, line 73: Call this a list. Calling it a set is a little misleading given that there is a literal set-type.

xaitk_saliency/utils/tools/drise_from_config.py
```diff
@@ -0,0 +1,130 @@
...
+
+        ref_dets_mat = dets_to_formatted_mat(ref_dets)
+
+        pert_dets_arrays = [dets_to_formatted_mat(pert_dets) for pert_dets in pert_dets_set]
+
+        # Use max detections found for a perturbed image as the number of proposals
+        num_props = max([array.shape[0] for array in pert_dets_arrays])
+
+        # Row of random numbers and negative objectiveness to pad proposals with
+        pad_prop = np.random.rand(ref_dets_mat.shape[1])
```
Purg, line 114: Random seems expensive for "empty" values that will not be considered. We could use `np.empty(<shape>)` here instead if truly any value will do. Otherwise `np.zeros` or `np.ones` is likely less expensive than uniform distribution random (I would assume).

xaitk_saliency/utils/tools/drise_from_config.py
```diff
@@ -0,0 +1,130 @@
...
+            [
+                bbox.min_vertex[0],
+                bbox.min_vertex[1],
+                bbox.max_vertex[0],
+                bbox.max_vertex[1]
+            ] for bbox in [det[0] for det in dets]
+        ])
+
+        class_mat = np.asarray([
+            list(class_dict.values())
```
Purg, line 60: It is not guaranteed (or it didn't use to be the case) that value enumerate is stable across different dictionaries. From 3.6 it is at least an implementation detail that dict keys/values are returned in insertion order, but we are not creating the dictionaries so we cannot actually know if they were all created uniformly. Basically, we need to cache a "class-key order" and then perform each dict-to-value-list cast w.r.t. that key-order.

xaitk_saliency/utils/tools/drise_from_config.py
```diff
@@ -0,0 +1,130 @@
...
+        # Row of random numbers and negative objectiveness to pad proposals with
+        pad_prop = np.random.rand(ref_dets_mat.shape[1])
+        pad_prop[4] = -1
+
+        for i, dets_array in enumerate(pert_dets_arrays):
+            num_pert_dets = dets_array.shape[0]
+
+            if num_pert_dets < num_props:
+                pert_dets_arrays[i] = np.vstack((
+                    dets_array.reshape(dets_array.shape[0], ref_dets_mat.shape[1]),
```
Purg, line 122: Why reshape? I would have thought that the matrices that come out of `dets_to_formatted_mat` would be have the same number of columns.

xaitk_saliency/utils/bin/drise_on_img_dir.py
```diff
@@ -0,0 +1,61 @@
...
+def drise_on_img_dir(img_dir, output_dir, config_file, perturbations, overlay_image, verbose):  # type: ignore
```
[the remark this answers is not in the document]
schencej: To be honest I don't really remember why I added that comment... but I will add annotations and a doc-string here.

xaitk_saliency/utils/tools/drise_from_config.py
schencej: I reshape here for anytime when `dets_array` is an empty and doesn't have a second dimension. Is there a better way to do this?

xaitk_saliency/utils/tools/gen_detector_sal.py
```diff
@@ -0,0 +1,137 @@
...
+        DRISE algorithm, as returned by the format_detection() function.
+        """
+
+        bbox_mat = np.asarray([
+            [
+                bbox.min_vertex[0],
+                bbox.min_vertex[1],
+                bbox.max_vertex[0],
+                bbox.max_vertex[1]
+            ] for bbox in [det[0] for det in dets]
```
Purg, line 58: If `dets` is a forward iterator (e.g. a generator) we will not be able to iterate over it twice here. In this use-case, since we're expanding it into flat structures anyway, it might as well be a list, so a first here may easily be `dets = list(dets)`.

tests/utils/test_gen_detector_sal.py
```diff
@@ -0,0 +1,191 @@
...
+            """
+            def detect_objects(
+                self,
+                img_iter: Iterable[np.ndarray]
+            ) -> Iterable[Iterable[Tuple[AxisAlignedBoundingBox, Dict[Hashable, float]]]]:
+                for i, img in enumerate(img_iter):
+                    yield [(AxisAlignedBoundingBox((0, 0), (100, 100)), {0: 1.0})] * i
+
+            def get_config(self) -> dict:
+                return {}
```
Purg, line 30: Since this and other similar methods defined here are not used, and we don't expect them to be, we can define here as
```
def get_config(self) -> dict: ...
```

I might also suggest defining them outside of the test method for code clarify reasons.

This change would address the codecov "failure" where the tests are not themselves being "covered" 100%.

xaitk_saliency/utils/tools/gen_detector_sal.py
```diff
@@ -0,0 +1,137 @@
...
+    :param img_perturber: Implementation of ``PerturbImage`` to use.
+    :param sal_generator: Implementation of
+        ``GenerateDetectorProposalSaliency`` to use.
+    :param verbose: Option to print out progress statements. Default is false.
+
+    :returns: List of saliency map matrices, each with shape
+        [nDets, imgHeight, imgWidth] respective to their corresponding image.
+    """
+
+    def dets_to_formatted_mat(
```
Purg, line 42: Does this method need to be defined in-line? I don't see it using any parent-scope values. If not, it would be more appropriate to define this method on the module scope, maybe giving it a prefix `_` to mark it as "private". This may also help with the complexity code-smell that SonarCloud is attributing to the parent function.

xaitk_saliency/utils/tools/gen_detector_sal.py
```diff
@@ -0,0 +1,137 @@
...
+        class_dicts = [det[1] for det in dets]
+
+        class_mat = np.asarray([
+            [class_dict[key] for key in class_keys]
+            for class_dict in class_dicts
+        ])
+
+        return format_detection(bbox_mat, class_mat)
+
+    if verbose:
```
Purg, line 73: We can reduce the quantity if `if verbose` lines here by assigning to a lambda a wrapper around the logging function, or a function that drops the input message on the floor:
```
if verbose:
    _log = LOG.info
else:
    _log = lambda _: None
```

We should also be using logging here instead of prints, e.g. create on the module level:
```
import logging
LOG = logging.getLogger(__name__)
```

Purg (commented):
Take a look at the SonarCloud and LGTM scan results to resolve some code-smells that should be relatively low-bar.

xaitk_saliency/utils/tools/gen_detector_sal.py
```diff
@@ -0,0 +1,137 @@
...
+        print("Getting reference detections.")
+
+    ref_dets_iter = detector(ref_imgs)
+
+    # List of saliency maps for all images
+    sal_maps_list = []
+
+    for i, (ref_img, ref_dets) in enumerate(zip(ref_imgs, ref_dets_iter)):
+
+        num_dets = sum(1 for det in ref_dets)
```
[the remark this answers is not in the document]
schencej: I iterate through `ref_dets` again later to get the class labels:
```
class_dicts = [det[1] for det in ref_dets] 
class_keys = list(class_dicts[0].keys())
```
So I think it would be most simple to just make it a list

xaitk_saliency/utils/tools/gen_detector_sal.py
```diff
@@ -0,0 +1,148 @@
...
+
+from smqtk_detection.interfaces.detect_image_objects import DetectImageObjects
+from smqtk_detection.utils.bbox import AxisAlignedBoundingBox
+
+from xaitk_saliency.interfaces.gen_detector_prop_sal import GenerateDetectorProposalSaliency
+from xaitk_saliency.interfaces.perturb_image import PerturbImage
+from xaitk_saliency.utils.masking import occlude_image_batch
+from xaitk_saliency.utils.detection import format_detection
+
+logging.basicConfig(level = logging.INFO)
```
Purg, line 13: Avoid initializing logging in the module, especially at the raw module level. This is something that should occur on a per-tool basis as it applies to that tool. E.g. any adding of handlers to a logger (which is what this does) should be in the CLI logic.

xaitk_saliency/utils/bin/det_sal_on_img_dir.py
```diff
@@ -0,0 +1,118 @@
...
+        and GenerateDetectorProposalSaliency implementations to use.
+
+    \f
+    :param img_dir: Directory containing the images.
+    :param output_dir: Directory to write the saliency maps to.
+    :param config_file: Config file specifying the ``DetectImageObjects``,
+        ``PerturbImage``, and ``GenerateDetectorProposalSaliency``
+        implementations to use.
+    :param overlay_image: Overlay saliency maps on images. Default is to output
+        write out saliency maps by themselves.
```
Purg, line 50: We should add here that when this is true we gray-scale the image if needed before overlaying a heatmap.

tests/utils/test_det_sal_on_img_dir.py
```diff
@@ -0,0 +1,236 @@
...
+            img.save(str(img_dir.join(f"test_{i}.png")))
+
+        config_file = tmpdir.join('conf.txt')
+
+        with config_file.open('w') as outfile:
+            json.dump(TEST_CONFIG, outfile)
+
+        output_dir = tmpdir.join('out')
+
+        output_config = tmpdir.join('config.txt')
```
Purg, line 89: There is a code-smell warning about using this string many times. I am going to flag it as a false-alarm in this case because this is just unit testing.

xaitk_saliency/utils/bin/det_sal_on_img_dir.py
```diff
@@ -0,0 +1,123 @@
...
+            plt.axis('off')
+            if overlay_image:
+                gray_img = np.asarray(Image.fromarray(ref_img).convert("L"))
+                plt.imshow(gray_img, alpha=0.7, cmap='gray')
+                plt.imshow(sal_map, cmap='jet', alpha=0.3)
+                plt.colorbar()
+            else:
+                plt.imshow(sal_map, cmap='jet')
+                plt.colorbar()
+            plt.savefig(os.path.join(sub_dir, f"det_{det_i}.jpeg"), bbox_inches='tight')
```
Purg, line 122: I just noticed running this locally that we are not fixing the colorbar legend scale, e.g. its outputting the min-max range per-detection-sal-map, which changes almost every time. My first instinct is that we should fix the range of the color bar across all outputs to the documented output standard of [-1, 1].

@brianhhu What are your thoughts on this?

Should we maybe get this in first, do a little session of "let's try using this" and then generate derivative issues/tasks from that, instead of continuing to delay the initial introduction of the tool?

brianhhu: Regarding the colorbar, my experience with RISE (at least the unbiased form, which will be true of D-RISE) is that saliency values tend to get squashed near the upper-end. So normalizing to [-1, 1] might result in pretty low-contrast, reddish saliency.

Regarding the PR itself, I think it makes sense to try to get this is in so we can do some further integration testing. We will probably have to make edits and fixes anyways as our use case evolves and we get more direction from the JAIC. Thanks!

xaitk_saliency/utils/tools/gen_detector_sal.py
```diff
@@ -0,0 +1,152 @@
...
+        _log(f'[{i+1}/{len(ref_imgs)}] Getting perturbation detections.')
+
+        # Detections on all perturbed images
+        pert_dets_iter = detector(pert_imgs)
+
+        _log(f'[{i+1}/{len(ref_imgs)}] Generating saliency maps.')
+
+        # use keys of first reference detection as keys for rest of detections
+        class_dicts = [det[1] for det in ref_dets]
+        class_keys = list(class_dicts[0].keys())
```
Purg, line 91: @schencej We were discussing last week how this could result in different key ordering from loop to loop, which would result in a bunch of detection matrices whose class confidence columns are not consistent/cross-applicable.
We should cache a `class_keys` across loops to mitigate this:
```
class_keys = None
for i, (ref_img, ref_dets) in enumerate(zip(ref_imgs, ref_dets_iter)):
    ...
    if class_keys is None:
        # class ordering based on first detection observed.
        # indexing into first detection -> classification dict (second position of tuple)
        class_keys = list(ref_dets[0][1].keys())
    # optionally error if any detection deviates from this set of keys
    # (something more performant than this)
    # if set(class_keys) != set(ref_dets[0][1].keys()):
    #     raise RuntimeError(...)
    ...
```

Purg: You were correct @schencej and I was incorrect in my reading: the use of the perturbed detections are local to the loop and are not collected in the parent scope, so we do not need to align the keys across loops. I apologize for the confusion!

Purg (approved):
This first tool iteration LGTM and am happy to merge if there aren't other objections.

Purg:
> This first tool iteration LGTM and am happy to merge if there aren't other objections.

Modulo autosquash rebase to resolve fixup commits.

PR #82, opened 2021-12-06
Add docker containerization

schencej:


.dockerignore
```diff
@@ -0,0 +1,9 @@
+codecov.yml
```
Purg, line 1: Most of the things here are not large and don't really impact the time it takes to create the build context except for "examples" if we've run a chuck of those locally (e.g. my sub-directory is >600MB). Since we are also using the repo root as the build context, `.git` should also be ignored. The content of this file could likely just be
```
/.git/
/examples/
```

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+RUN pip install poetry
+RUN poetry config virtualenvs.create false
+  
+WORKDIR /xaitk
+
+COPY poetry.lock pyproject.toml /xaitk/
+RUN poetry install --no-root
+  
+COPY . /xaitk/
+RUN poetry install
```
Purg, line 12: I assume that you have these separated so that updates to the XAITK package do not generally necessitate a full dep install when rebuilding. If so, could you comment as such here so that it is made clear?

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM python:3.8 AS xaitk_base
+
+# install poetry and configure to install to pip environment
+RUN pip install poetry
+RUN poetry config virtualenvs.create false
```
Purg, line 5: These can be collapsed into the same build layer (good practice, reduces result image size)
```suggestion
RUN pip install poetry \
 && poetry config virtualenvs.create false
```

Purg (approved):
LGTM. Please `git rebase origin/master -i --autosquash` and we can get this merged.

PR #83, opened 2021-12-07
Fix bug in D-RISE Scoring implementation

brianhhu:
* Fix cosine similarity computation
* Remove unneeded mask inversion


xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py
```diff
@@ -108,9 +108,9 @@ def generate(
                       ref_dets[:, :4]).reshape(n_masks, n_props, n_dets)
 
         # Compute similarity of class probabilities
-        s2 = cdist(perturbed_dets[:, :, 5:].reshape(n_masks * n_props, -1),
-                   ref_dets[:, 5:],
-                   metric=self.proximity_metric).reshape(n_masks, n_props, n_dets)
+        s2 = 1 - cdist(perturbed_dets[:, :, 5:].reshape(n_masks * n_props, -1),
+                       ref_dets[:, 5:],
+                       metric=self.proximity_metric).reshape(n_masks, n_props, n_dets)
```
brianhhu, line 113: One issue here is that this change to the `1 - cdist(...)` operation may no longer be valid for other non-cosine proxity metrics, e.g. Euclidean distance. I'm not sure how to best address this, may be with a warning in the `__init__` block?

Purg: If we only know D-RISE to "work" with cosine similarity, should we then just take out the option to configure to other things?

examples/DRISE.ipynb
```diff
@@ -483,13 +483,13 @@
      "text": [
       "Pert preds: (200, 100, 85)\n",
       "Saliency maps: (2, 480, 640)\n",
-      "Class 1 saliency map range: [0.8693522490050113, 1.0]\n",
-      "Class 2 saliency map range: [0.856367198172182, 1.0]\n"
+      "Class 1 saliency map range: [0.9604193453572968, 1.0]\n",
+      "Class 2 saliency map range: [0.9703758275703027, 1.0]\n"
```
Purg, line 487: Now the saliency score range is even more densely packed into the upper "positively salient" range. Is this of any specific concern?

xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py
```diff
@@ -108,9 +108,9 @@ def generate(
                       ref_dets[:, :4]).reshape(n_masks, n_props, n_dets)
 
         # Compute similarity of class probabilities
-        s2 = cdist(perturbed_dets[:, :, 5:].reshape(n_masks * n_props, -1),
-                   ref_dets[:, 5:],
-                   metric=self.proximity_metric).reshape(n_masks, n_props, n_dets)
+        s2 = 1 - cdist(perturbed_dets[:, :, 5:].reshape(n_masks * n_props, -1),
```
Purg, line 111: The `1 - ` component of this may want to be replaced with a uint8 1 values to not up-case the `cdist` results in the same way that was happening in the `weight_regions_by_scalar` mask inversion.

examples/DRISE.ipynb
brianhhu: I'm not 100% sure about this, but I think RISE/D-RISE are not well calibrated, which was the motivation for a de-biased form of RISE. In my preliminary testing with D-RISE, increasing the number of masks from 200->400->800 seemed to give "better" saliency results (but also squished the saliency range values further).

image

xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py
brianhhu: This is kind of a philosophical question- we also use the `cdist` operation in our `SimilarityScoring` implementation, and from what I can tell, the original implementation mainly tested it with Euclidean distance, yet we currently make available the whole gamut of distance metrics, many of which may not be valid for that implementation.

xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py
brianhhu: It looks like the `cdist` function automatically upcasts returned results (see example below).

```
from scipy.spatial.distance import cdist
coords = np.array([(35.0456, -85.2672),
          (35.1174, -89.9711),
          (35.9728, -83.9422),
          (36.1667, -86.7833)]).astype('float32')
print(cdist(coords, coords, 'euclidean').dtype)  # should be 'float64'
```

Should we just convert the `cdist` results to `float32`?

xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py
Purg: It could be the case that we over-engineered that impl as well, however it seems that works on distance (metric?) functions, which all the cdist options are I thought. It's also OK if we "fix" this impl while the other one is still under question.

xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py
Purg: If `cdist` is doing unknown casting and we want to make sure that out math stays within known type constraints, then yes. However, instead of just always casting to a constant, we should use `np.promote_types(T1, T2)` to determine the cast type at runtime, e.g.
```ipython
In [5]: np.promote_types(np.uint8, np.uint16)
Out[5]: dtype('uint16')
```

xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py
```diff
@@ -124,7 +124,9 @@ def generate(
         s = s.max(axis=1)
 
         # Weighting perturbed regions by similarity
-        sal = weight_regions_by_scalar(s, perturbed_masks)
+        sal = weight_regions_by_scalar(s,
+                                       perturbed_masks,
+                                       inv_masks=False)
```
nitesh201, line 129: out of curiousity, is there a reason that inv_masks is "True" by default?

xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py
nitesh201: I agree that, if the impl only really "works" w/ a specific distance metric, than perhaps it's better not to parametrize this

xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py
nitesh201, line 129: out of curiousity, is there a reason that inv_masks is "True" by default?

nitesh201 (approved):

xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py
brianhhu: Probably a historical choice, the first algorithm to use that was the `OcclusionScoring` implementation (with sliding windows), which required a mask inversion when computing saliency. We could discuss changing the default behavior now that this function is used in multiple places.

Purg (approved):
LGTM. I'm happy to merge if no one has any objections.

nitesh201 (approved):

nitesh201:
same here - I would like to revisit the question of whether we should be inverting masks by default at some point, but agree that it's not worth a potentially breaking API change until this is understood

PR #85, opened 2022-01-08
Add tool to generate saliency for detections in COCO dataset

schencej:


schencej:
SoundCloud fails due to the same security concern with the logger that occurred with the previous tool.

Purg:
Unit tests please.

Purg:
The coverage on the test files should be 100%. Conditional logic like the is_usable flag should be encapsulated in the library where possible (which I believe it can be in this case, e.g. import kwcoco from the gen_coco_sal). We should also try to hit that 80% mark on coverage to make sure we're hitting the majority of functionality within the tests.

tests/utils/test_sal_on_coco_dets.py
```diff
@@ -0,0 +1,102 @@
...
+config_file = os.path.join(DATA_DIR, 'config.json')
+
+
+class TestSalOnCocoDets:
+    """
+    These tests make use of the `tmpdir` fixture from `pytest`. Find more
+    information here: https://docs.pytest.org/en/6.2.x/tmpdir.html
+    """
+
+    @pytest.mark.skipif(not usable, reason="Extra 'xaitk-saliency[tools]' not installed.")
```
Purg, line 25: If this is on all of the test functions, this decorator can be applied to the class to bulk skip.

tests/utils/test_sal_on_coco_dets.py
```diff
@@ -0,0 +1,102 @@
...
+import os
+import py
+import pytest
+
+from tests import DATA_DIR
+
+from xaitk_saliency.utils.bin.sal_on_coco_dets import sal_on_coco_dets
+
+from importlib.util import find_spec
+deps = ['kwcoco']
```
Purg, line 11: Logic should have some whitespace separation from the import statements.

pyproject.toml
```diff
@@ -73,6 +76,7 @@ ipython = "^7.22.0"
 
 [tool.poetry.scripts]
 det-sal-on-img-dir = "xaitk_saliency.utils.bin.det_sal_on_img_dir:det_sal_on_img_dir"
```
Purg, line 78: With the removal of the defunct tool we need to remove this tool script export as well.

Purg:
Given your sample detection COCO json, I was able to successfully generate a config file, tweak it, and run this using the patched CenterNet plugin from your parallel branch of smqtk-detectors. For some configurations of this, RAM blows up at the saliency map generation part, but that's not an issue introduced by this tool. Maybe another iteration of this could include batching `ref_dets_mat` when generating saliency per image, but lets have that be a separate experiment/branch.

brianhhu:
Just out of curiosity, what was the order of magnitude for `ref_dets_mat`? How many reference detections were we trying to explain? We've been putting off re-designing the D-RISE algorithm, but it does seem to be a bit memory-inefficient due to the use of alarge matrix multiplication.

xaitk_saliency/utils/gen_coco_sal.py
```diff
@@ -0,0 +1,262 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+# mute logger to suppress verbose messages
+def _mute_log(_: Any) -> None: ...
+
+
+if not is_usable:
+    print("This tool requires additional dependencies, please install 'xaitk-saliency[tools]'")
```
Purg, line 29: This should use `LOG.warning` or `warnings.warn` instead of a print. That level of logging should always be printed.

We should also be more specific than "this tool" and more specifically refer to the `__name__` of this module. That way the user knows where this is coming from (though I suppose `warnings.warn` would also denote this location).

xaitk_saliency/utils/gen_coco_sal.py
```diff
@@ -0,0 +1,262 @@
...
+            found.
+        """
+        for cat_id in cats:
+            if name == cats[cat_id]['name']:
+                return cat_id
+
+        return -1
+
+    def _map_score_dict_keys(
+        score_keys: KeysView,
```
Purg, line 129: This could maybe be a more general `Iterable[Hashable]` type, but this likely works for now until this private method gets used elsewhere.

xaitk_saliency/utils/gen_coco_sal.py
```diff
@@ -0,0 +1,262 @@
...
+
+        :param name: Category name.
+        :param cats: Dictionary of categories, as taken from the "categories"
+            section of a COCO style annotation dictionary.
+        :returns: Id of matching category, or -1 if no matching category is
+            found.
+        """
+        for cat_id in cats:
+            if name == cats[cat_id]['name']:
+                return cat_id
```
Purg, line 124: Accesses (read: "expensive" function calls) into the `cats` dict could be reduced a little by using:
```suggestion
        for cat_id, cat_obj in cats.items():
            if name == cat_obj['name']:
                return cat_id
```

xaitk_saliency/utils/gen_coco_sal.py
```diff
@@ -0,0 +1,262 @@
...
+            section of a COCO style annotation dictionary.
+        :returns: Dictionary mapping score dict key to category id.
+        """
+
+        cat_map = {}  # type: Dict[Hashable, int]
+        for key in score_keys:
+            if type(key) is int or type(key) is float or (type(key) is str and key.isnumeric()):
+                # assume key is a cat id
+                if int(key) in coco_cats:
+                    cat_map[key] = coco_cats[key]['id']
```
Purg, line 148: 2 things:
1) I think this will fail if `type(key) is not int`. You are casting `key` to int before checking containment and then accessing with the original type.
2) Wouldn't `coco_cats[key]['id']` be the same as `int(key)`? If so, the assignment could just be of `int(key)` to save on total function calls:
```suggestion
                key_int = int(key)
                if key_int in coco_cats:
                    cat_map[key] = key_int
```

xaitk_saliency/utils/gen_coco_sal.py
```diff
@@ -0,0 +1,262 @@
...
+                else:
+                    raise MismatchedLabelsError("Provided detections and detector have mismatched class labels.")
+            else:
+                # try and match key with category name
+                cat_id = _name_to_cat_id(key, coco_cats)
+                if(cat_id == -1):
+                    raise MismatchedLabelsError("Provided detections and detector have mismatched class labels.")
+                cat_map[key] = cat_id
```
Purg, line 156: I recall that you asked for some specific feedback on this section. I think this is OK as is, just as long as "id" types that `key` may take is only what is checked on L145.

A possibly "better" approach here, that is more akin to duck-typing, is a try-except. E.g. try the more specific/faster access and if that fails try the broader access. In this context, I see you are casting `key` to an integer in the top-if-condition, so maybe:
```suggestion
        for key in score_keys:
            try:
                # assume key is a cat id
                key_int = int(key)
                if key_int in coco_cats:
                    cat_map[key] = coco_cats[key]['id']  # assign just key_int?
                else:
                    raise MismatchedLabelsError("Provided detections and detector have mismatched class labels.")
            except ValueError:  # failed integer cast
                # try and match key with category name
                cat_id = _name_to_cat_id(key, coco_cats)
                if(cat_id == -1):
                    raise MismatchedLabelsError("Provided detections and detector have mismatched class labels.")
                cat_map[key] = cat_id
```

Either way, things will get weird if a data-set's category names *are* numbers. I don't have a solution in mind for that case, and I don't know if we should address that yet/here.

xaitk_saliency/utils/bin/sal_on_coco_dets.py
```diff
@@ -94,37 +99,52 @@ def det_sal_on_img_dir(
...
+        sal_maps = gen_coco_sal(
+            dets_dset,
+            blackbox_detector,
+            img_perturber,
+            sal_generator,
+            verbose=verbose
+        )
+    except MismatchedLabelsError as e:
+        print(str(e))
+        exit()
```
Purg, line 112: Why not just let python's exception handling print the stack-trace and error message for us? This exit will return a 0-code which can mislead callers to thinking the tool exited "successfully."

xaitk_saliency/utils/bin/sal_on_coco_dets.py
```diff
@@ -69,21 +78,17 @@ def det_sal_on_img_dir(
 
         exit()
 
-    files = os.listdir(img_dir)
-
-    img_files = []
-    imgs = []
+    if not is_usable:
+        print("This tool requires additional dependencies, please install 'xaitk-saliency[tools]'")
+        exit()
```
Purg, line 83: This should likely not exit with a 0 return code as this is not a successful condition.

Purg (approved):
LGTM. Please rebase-autosquash and we can get this in.

PR #86, opened 2022-01-19
Add example notebook generating saliency maps for serialized detections

schencej:
This is the first pass at our demo notebook. The notebook sets up the environment and demonstrates how to use the `gen_coco_sal()` function.


schencej:
Marking as draft until #85 is merged.

Purg:
Just had the thought that this notebook should be added to `.github/workflows/ci-example-notebooks.yml` if it makes sense to run this at CI time (which should default to "we should" unless we have a reason not to).

Purg (commented):
I apparently cannot comment inline with the notebook due to being too large for github to want to render, so I'll bullet them here:
* "... which includes the dependencies for the CenterNetVisdrone class" -> "which includes the dependencies for the pytorch-based CenterNetVisdrone plugin"
* I think we should bring up / adapt the introduction sentence that's under the "Load Detections and Visualize" section into the "Download Detections and Image" section.
* The "Load Detections and Visualize" section could provide a brief introduction and reference to kwcoco, i.e. that we're using it as a provider of robust load and query tools for COCO-format inputs. (or something along those lines)
* In "Instantiate Detector" we may want to provide a construction with `cuda=False`. Those that have GPUs may opt into using it by changing this parameter, which we could highlight in the section description.

Purg:
This notebook also likely needs a
```
# Use JPEG format for inline visualizations here.
%config InlineBackend.figure_format = "jpeg"
```
line in one of the earlier cells to bring down the file-size. Ghostflow currently checks for and errors if a single file is classified as a BLOB and expands above 1MB in size as means of protecting against compounding repo object bloat. We should be able to bring this size way down by changing the image generation from outputting lossless PNG images to instead generate compressed JPEG images.

schencej:
I addressed all of your comments I believe plus added a more comprehensive intro that talks about the function in more depth. I also rearranged to instantiate the detector first before loading the detections so that it would be more clear that they are predictions made by the CenterNet model.

Purg:
LGTM. @brianhhu feel free to approve when you take your look. Of course, autosquash rebase at that time.

brianhhu:
Overall, LGTM- thanks! To help the reader (and assuming it's useful here), we could add links to the [COCO annotation format](https://cocodataset.org/#format-data), [CenterNet model](https://arxiv.org/pdf/1904.07850.pdf), and [VisDrone dataset](https://arxiv.org/pdf/1804.07437v2.pdf) when referenced in the notebook, similar to what we did for the Colab notebook. A few minor nitpicks for spelling and compliance with our [style sheet](https://github.com/XAITK/xaitk-saliency/blob/master/docs/misc/style_sheet.rst) (feel free to ignore):

1) programatic - > programmatic
2) detectors output -> detector's output
3) pytorch-based -> PyTorch-based
4) DRISE -> D-RISE

schencej:
Added two additional lines to the environmental setup to uninstall opencv and install the headless version. This allowed it to pass CI.

Purg:
I just notices / thought about it: we are missing a cell with Google Colab link, analogous to the other notebooks (documented in our CONTRIBUTING.md).

schencej:
Colab link added.

Purg (approved):
Sorry for the incremental finds, LGTM now.

brianhhu:
LGTM, thanks!

PR #88, opened 2022-01-26
Fix log message for gen_coco_sal and remove gen_detector_sal release note

schencej:
The log messages for `gen_coco_sal` prints the current image that saliency is being calculated for out of the total number of images, "(2 / 10)" for example if it was on the 2nd out of 10. The total number of images was previously calculated as the number of images in the kwCoco dataset, leading to an incorrect value if there are images that do not have any detections associated with them. This didn't affect the saliency maps as images with no detections were skipped. I changed the behavior a bit to fix the log message.

Also I noticed that the `gen_detector_sal` module that we removed is still in the release notes, so I removed it.

### DIFF
--- a/docs/release_notes/pending_release.rst
+++ b/docs/release_notes/pending_release.rst
@@ -36,12 +36,6 @@ Implementations
 
 Utilities
 
-* Add ``gen_detector_sal`` function to compute saliency maps for a detection
-  algorithm using any chosen implementations of ``DetectImageObjects``,
-  ``PerturbImage``, and ``GenerateDetectorProposalSaliency``, with accompanying
-  cli script ``det-sal-on-img-dir`` that takes a config file for those
-  implementations.
-
 * Add ``gen_coco_sal`` function to compute saliency maps for detections in a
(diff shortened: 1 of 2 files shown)


docs/release_notes/pending_release.rst
```diff
@@ -85,3 +79,7 @@ Utilities
 
 * Fix ``xaitk_saliency.utils.masking.weight_regions_by_scalar`` to not upcast
   the data type output when ``inv_masks is True``.
+
+* Fix ``xaitk_saliency.utils.gen_coco_sal.gen_coco_sal`` to filter out images
+  in the COCO dataset that do not have any detections associated with them.
+  This fixes the info log to print the correct number of images.
```
Purg, line 85: I don't think we need to add this "fix" note since we just introduced the tool within this same "release." I.e. it's a little odd to introduce that we added `gen_coco_sal` (the new L39 above) and then fixed the very same thing.

schencej:
Amended the first commit to remove the addition to `pending_release.rst`.

Purg (approved):

PR #89, opened 2022-01-29
Add example using saliency to explain edge case detections.

schencej:
First pass of edge case saliency example. My system has limited resources for saliency map generation and I haven't included any discussion on the insight that the maps provide, but please take a look and provide feedback.


schencej:
`kwplot` is a dependency for the `show_image()` function from `kwcoco` which is used here. A pip install could be added to the environmental setup cell. The CI run fails without this. Opinions?

Purg:
> `kwplot` is a dependency for the `show_image()` function from `kwcoco` which is used here. A pip install could be added to the environmental setup cell. The CI run fails without this. Opinions?

I think it would be OK to add a notebook-local pip-install of `kwplot` in this case, since this is first place it is being utilized. I'm open to other thoughts. I am hesitant to jump into requiring it as part of the extras dependencies until we observe more utilization.

brianhhu:
^ That sounds reasonable to me. I'm still not 100% sure how I feel about separating the initial presentation of example model failures with their corresponding saliency maps, as that seems to introduce a gap there which might be hard for a reader to bridge (e.g. scrolling back to see the exact model misclassification and how that relates to the saliency map). If we do decide to show results and saliency maps side-by-side, we might be able to make use of default matplotlib functionality for viewing detections, which would remove the need to use the functionality in kwplot.

examples/EdgeCaseSaliency.ipynb
```diff
@@ -0,0 +1,1318 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "47395fb4",
+   "metadata": {},
+   "source": [
+    "# Saliency to Explain Edge Cases\n",
+    "\n",
+    "This demo demonstrates the use of the `xaitk_saliency` package to provide useful explanations for the performance of an object detection model.\n",
+    "Specifically, we examine instances where the decisions the model makes are questionable.\n",
+    "Saliency maps can provide useful insights into why the model behaved in an unexpected way."
```
Purg: We should more specifically introduce the use-case right off the bat. This by itself reads a too generic to me and doesn't really give me an idea of what the rest of the notebook/demo is going to entail. We should treat this first cell like a paper abstract and introduce a complete sketch of what will follow.

examples/EdgeCaseSaliency.ipynb
```diff
@@ -0,0 +1,1318 @@
...
...FsKx75HQ0KJcakom9cWdxbTGKeJ43HZxio38vyXibOGGM9x7irj+I5dXRIdVRZSOYrq1GWAP95O4+lOvtA1G2BKxCePs0RyfxHUflU+p1QqRmrM5+SGRflWaIg924NNj0/cCfkkI/u4P8qdcRyRMRIjofRwR/OprWIou7oTRypbCdLm2ZSceZMIU7cHHb1qyYyI9qHacYHGcVdypYsyKzHq2MH86gkQ5JiAP+yxwfz6VNmiasJaJLRGPLa3zMwaSNlPTjBNe8fBWLyfBl0uAD9vcnH/XOOvHHyiL5q7GPY817R8Hv+RSu/T7e/8A6Ljpx3MG3ax6FRRRWhBm+IQD4a1UHp9jm/8AQDXgDQogaVgdo9a9/wDEIz4a1UAZzZzf+gGvnzypbh9ikEDk88Cs6gDixkH7tQifmTTRvXAViGPTmpDbPHHtV/3mc44GBUf2dY42d28yRjhR2H4VmMRlaPCBssT+A96upF5EYjI+c/M/17D8v51BbxZYlgWRPmc46eg/Gp3c/eJ+Y8nFb0Y68xjVl0M7VL2OztyWIL/wj1rkVdpJScksSSTirmuXPnaiyDJVBjHvVaEc4LOPTAqqjMokyRKxG5yGHXOKtLGIFBA3sR61GAGbapOW4weBU6WjOdqyKD3J6Vg2ax20I3LIvzKcEZYqOtPikzMwj3BAhzgE9qkd1gQISrbMZb1qAXSvgJHt+Ug7e/FJO62IWobWuQNkZYBu46dKuTW/nwouVaUYBUDOaSCNhacEqrkEgcHH86uWkckNpLMMbcjaPboacexajYymSbIiJ2qDhR/hUF0jQIzucuOma0ltz9oiVjk7s8VV1CJJ7lovmBJAwOc/4U0OSsUIZJ/LJB+duy9BV20NxM2x1UsejHjFXbKzNuokbau7+A8tj8elWLkIsO2N9iAndt9+1ctSrrZGUmc/KCkhRjt5x8vOaqPI6srbC+0YBY8j8q1ZrOFYl5OCclQcnFLDaQyFV8v5uw254+pNWqisK5jo8jEtIxGBwPQVsaZJhyEZckZGe/8AnNK9tZRSGMq8ZZuW6nH51NDbpHMpjjYIBjMnf3FUqqZUZK50kHzRmnpgNiorZwVxnqKfn5664vQ2kWww4xTlPzGolPFSA/MOe1UIlqXh4iDUQNSR9xQ9iVuZOqaRDfRFmjVpF+6SKq2Wlm1GY4kxnj1rf4xg1GFP15rB
```
Purg: We will likely want provide a "saved" notebook with this set to False in order to add the lease amount of "distraction" (read: logging statements) as possible.

Purg:
Related to plotting, I would also add that we don't have any legends present on the plots, which might reinforce @brianhhu's mention of just using `matplotlib` directly, unless `kwplot` can do all the stuff mentioned with different parameterization.

PR #91, opened 2022-02-07
Update weight_regions_by_scalar to not use matrix multiplication

schencej:
From my testing, completely removing matrix multiplication reduced both execution time and memory allocated significantly. Therefore I opted to update the implementation of `weight_regions_by_scalar()` instead of adding another version.

On my machine, using 2000 masks and 10 classes, the old implementation took ~26s and allocated ~57G of memory. This update reduces that to ~8s and ~13G.


xaitk_saliency/utils/masking.py
```diff
@@ -275,21 +275,31 @@ def weight_regions_by_scalar(
...
-    # Aggregate scores across all perturbed regions.
-    sal_across_masks = np.transpose(heatmap.sum(axis=2))
+    # split weights per class
+    for i, class_scales in enumerate(np.transpose(scalar_vec)):
+        # aggregate scaled masks
+        for (mask, scalar) in zip(masks, class_scales):
+            sal_across_masks[i] += mask * scalar
 
     if normalize:
-        # Removing regions that are never masked to avoid a dividebyzero warning
```
brianhhu, line 288: Any reason to remove the comments here? I would err on the side of being a bit more verbose in case a reader has questions about why this normalize operation is included here.

schencej: I think I had some copy-paste errors. Re-added the comments in that section.

schencej:
I've done some further investigation, and I believe the vectorized computation is slower due to the amount of memory needed for it. The added overhead of reading and writing to memory slows it down.

The plot below shows execution time of the new version vs the legacy (vectorized) version as a function of mask size. For smaller images, the old version is faster, but the execution time grows seemingly exponentially as the size increases.

image

This will obviously vary system-to-system, but given the average size of masks that we use is pretty high, the new version  makes more sense IMO.

docs/release_notes/pending_release.rst
```diff
@@ -82,3 +82,6 @@ Utilities
 
 * Fix ``xaitk_saliency.utils.masking.weight_regions_by_scalar`` to not upcast
   the data type output when ``inv_masks is True``.
+
+* Update ``xaitk_saliency.utils.masking.weight_regions_by_scalar`` to not use
+  matrix multiplication which significantly improve efficiency.
```
brianhhu, line 87: So, technically the function still uses matrix multiplication (i.e. the `mask * scalar` operation)- maybe instead:

```suggestion
* Update ``xaitk_saliency.utils.masking.weight_regions_by_scalar`` to not use
  fully vectorized operation which significantly improves efficiency.
```

brianhhu: And a SO thread that might be somewhat relevant: https://stackoverflow.com/questions/48887461/why-python-loops-over-slices-of-numpy-arrays-are-faster-than-fully-vectorized-op

brianhhu (approved):
Overall, LGTM- thanks for making this change, it has been on my wish list for some time!

PR #92, opened 2022-02-11
Add logic for two dim s value in RISE

schencej:


schencej:
Some input on how to resolve the typecheck errors would be appreciated.

xaitk_saliency/impls/perturb_image/rise.py
```diff
@@ -92,6 +102,18 @@ def perturb(
...
+        if type(s) is tuple:
+            s = np.array(s)
+
+        # Dynamic square masking
+        if grid is None:
+            val_idx = 1 - list(s).index(None)
+            r = input_size[val_idx] / s[val_idx]
+            s[1 - val_idx] = np.ceil(input_size[1 - val_idx] / r).astype('int')
+            grid: np.ndarray = np.random.default_rng(self.seed).random((self.n, *s)) < self.p1
+            grid = grid.astype('float32')
```
brianhhu, line 115: I'm a bit torn about this dynamic square masking, as I think it introduces extra complexity and may generate unexpected behavior (although I cannot think of an example right now off-hand). What are the drawbacks of asking the user to explicitly specify grid size in x/y, similar to how we also ask the user to specify sliding window size, stride, etc. in other perturbation approaches where we provide a tuple input?

@Purg Do you have any thoughts on this?

schencej: The drawback that I saw is when you are using the same image perturber on multiple images with different aspect ratios. Using a set (x,y) might yield different grid cell shapes for each image. The dynamic masking can't ensure each grid is same size but they are all square at least.

xaitk_saliency/impls/perturb_image/rise.py
```diff
@@ -78,8 +81,15 @@ def __init__(
...
-        grid = grid.astype('float32')
+        if type(s) is tuple:
+            if None in s:
+                grid = None  # type: Optional[np.ndarray]
+            else:
+                grid = np.random.default_rng(seed).random((n, *s)) < p1
+                grid = grid.astype('float32')
+        else:
+            grid = np.random.default_rng(seed).random((n, s, s)) < p1
+            grid = grid.astype('float32')
```
Purg, line 92: It would be more thematic to duck-typing to attempt to *-expand `s` and catch the `TypeError` raised when failing to expand a single integer, as opposed to specifically catching the tuple type.

Purg: I think this comment is defunct since I didn't realize a "dynamic" mode was being used in the tuple-with-a-None case.

xaitk_saliency/impls/perturb_image/rise.py
Purg: We may be entering the territory of maybe wanting a different perturber impl. On one level, the RISE perturber is doing what the paper described it does, and we likely want to keep it like that to be reflective of that method. This was probably based on the assumption that imagery is already preprocessed to a "standard" size, or to the target-network's input data requirements. This is of course not robust or applicable to datasets like VisDrone with imagery coming from multiple sources.

What we might more ideally want is a new perturber impl like `SlidingWindow`, but with random window weights like RISE, but with more controlled fix-pixel or width/height percentage parameters so that mask regions are consistent in size/placement. Is that what you are trying to get at with this @schencej?

brianhhu: Related to the above, I was envisioning a use case where we still have a standard input size, but it may not be square (e.g. 640x480 or 1920x1080). In the case the user wants to have square-like grid cells, the current RISE approach cannot handle this since the `s` parameter is assumed to be the same along x/y (resulting in rectangular masking regions). Specifying a tuple allows the user to essentially adapt the grid cells to the aspect ratio (but again assuming this is fixed for a set of inputs).

schencej: Yes, my main goal was to keep the grid cells as similarly sized as possible across the different images I've been analyzing. I think specifying this at a pixel size level would be even better. Perhaps getting rid of the "dynamic" feature here and creating another implementation would make the most sense.

Purg:
Given #95, is this PR safe to close?

schencej:
> Given #95, is this PR safe to close?

Yes, go ahead.

PR #93, opened 2022-02-16
Update behavior of gen-coco-sal with one-hot encoded dets

schencej:
Replace "objectness" with score and one-hot encode scores when only one score is returned.

No note in `pending_release.rst` because `gen_coco_sal()` is already listed.


schencej:
I realized that we are only looking for the "score" key when reading the serialized detections. Therefore there will always only be one score for the reference detections. Going to work on this some more.

schencej:
Added logic for the `prob` annotation field which is a list of class probabilities. If this key is present in the serialized detection, the class scores are encoded normally. If the `score` key is only present, then it is one-hot encoded with the objectness replaced by the score.

xaitk_saliency/utils/gen_coco_sal.py
```diff
@@ -256,8 +262,17 @@ def _coco_dets_to_formatted_mat(
...
         max_cat_id = max(coco_dset.cats.keys())
         class_mat = np.zeros((len(dets), max_cat_id+1))
+        objectness = np.ones(len(dets))
 
         for i, det in enumerate(dets):
-            class_mat[i, det['category_id']] = det['score']
 
-        return format_detection(bbox_mat, class_mat)
+            # look for class probabilities
+            if 'prob' in det:
```
brianhhu, line 270: This seems a bit fragile- what if the user doesn't use the `prob` key to encode class probabilities? For example, it's conceivable that a user could still use the `score` key to store class probabilities, since the [official COCO documentation](https://cocodataset.org/#format-results) only mentions a `score` key. In this case, would it be safer to check the shape/length of the value instead of a specific key?

brianhhu:
Just had another thought- I think at one point, we had also discussed adding something to the `DRISEScoring` saliency map generation [docstring](https://github.com/XAITK/xaitk-saliency/blob/master/xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py#L10-L22) describing how to use the algorithm in the case where the detection model only outputs a single class and confidence (i.e. using confidence as objectness and one-hot encoding class scores), instead of using the full set of class probabilities. Could we add that as part of this PR too?

xaitk_saliency/utils/gen_coco_sal.py
schencej: `prob` is used by the [kwcoco schema](https://kwcoco.readthedocs.io/en/latest/_modules/kwcoco/coco_schema.html?highlight=prob), which is why I used it here. To me, it seems confusing to use `score` for the class probabilities given that the official COCO schema specifies it as float. Perhaps we could be more explicit in documentation about which key to use.

brianhhu: Ok, that makes more sense now- please disregard my previous comment then. Thanks for the clarification!

Purg:
> Just had another thought- I think at one point, we had also discussed adding something to the `DRISEScoring` [docstring](https://github.com/XAITK/xaitk-saliency/blob/master/xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py#L10-L22) describing how to use the algorithm in the case where the detection model only outputs a single class prediction and confidence (i.e. using confidence as objectness and one-hot encoding class scores), instead of using the full set of class probabilities. Could we add that as part of this PR too?

Is that advice is generation method agnostic? E.g. would that advice more pertain to the `xaitk_saliency.utils.detection.format_detection` function?

xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py
```diff
@@ -21,6 +21,11 @@ class DRISEScoring (GenerateDetectorProposalSaliency):
     bounding box locations, 1 index for objectness, and nClasses indices for
     different object classes).
 
+    If any of your detections consist of a single class prediction and
+    confidence score instead of scores for each class, it is best practice to
+    replace the objectness score with the confidence score and one-hot encode
+    the class scores.
```
brianhhu, line 27: Minor edits/suggestions for clarity:

```suggestion
    If your detections consist of a single class prediction and confidence score
    instead of scores for each class, it is best practice to replace the objectness
    score with the confidence score and use a one-hot encoding of the prediction
    as the class scores.
```

brianhhu (approved):
This LGTM, regarding Paul's comment about whether the comment should go in the `DRISEScoring` or `format_detections` docstring- I think it makes sense to put it in D-RISE, as it's unclear whether other detection-based saliency algorithms would make use of the same set of inputs. However, I'm open to discussing this more.

xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py
schencej: Updated.

brianhhu: Thanks!

brianhhu:
@Purg LGTM, I have already approved the PR. Feel free to merge (but will have to bypass the ghostflow check). Thanks!

PR #95, opened 2022-02-23
Add RandomGrid image perturber

schencej:


Purg:
Rebased on updated master

xaitk_saliency/impls/perturb_image/square_grid.py
```diff
@@ -0,0 +1,82 @@
...
+        s = np.array([self.s, self.s])
+        p = self.p
+        threads = self.threads
+        grid_rng = np.random.default_rng(self.seed)
+        shift_rng = np.random.default_rng(self.seed)
+
+        img_shape = np.array(ref_img.shape[:2])
+
+        grid_size = (img_shape // s) + 1
+        grid_size += img_shape % s != 0
```
Purg, line 38: Things seem to work, but this math threw me for a loop as to why the `+1` is here, etc. A comment or two might address why this is being done.

schencej:
- Added docstring and comments
- Changed the `s` parameter to be a tuple, requiring the user to specify the cell size along both dimensions. Also changed the class name to a more generic `RandomGrid` since it is not explicitly square now
 - If you used a cell size of 1, the max shift was 0 which caused an error to be thrown by the random number generator. I added logic for this

schencej:
Is mentioning RISE and/or including its license in the docstring something we should do here? The logic is similar.

xaitk_saliency/impls/perturb_image/random_grid.py
```diff
@@ -0,0 +1,118 @@
...
+        :param s: Dimensions of the grid cells in pixels. E.g. (3, 4) would use
+            a grid of 3x4 pixel cells.
+        :param p: Probability of a grid cell being set to 1 (not occluded).
+            This should be a float value in the [0, 1] range.
+        :param seed: A seed to use for the random number generator, allowing
+            for masks to be reproduced.
+        :param threads: Number of threads to use when generating masks. If this
+            is <=0 or None, no threading is used and processing is performed
+            in-line serially.
+        """
```
Purg, line 34: Somewhat minor, but please move this doc-string to the class level.

Purg:
> Is mentioning RISE and/or including its license in the docstring something we should do here? The logic is similar.

We could mention that this impl is based on RISE and exists to address the changing-"square"-size aspect of the original RISE implementation. I think it is important to capture where we came from (RISE) and why we are here (fixed pixel rectangles).

Purg (approved):
LGTM.

xaitk_saliency/impls/perturb_image/random_grid.py
```diff
@@ -0,0 +1,124 @@
...
+
+    This method is based on RISE (http://bmvc2018.org/contents/papers/1064.pdf)
+    but aims to address the changing cell size, given images of different
+    sizes, aspect of that implementation. This method keeps cell size constant
+    and instead adjusts the overall grid size for different sized images.
+
+    :param n: Number of masks to generate.
+    :param s: Dimensions of the grid cells in pixels. E.g. (3, 4) would use
+        a grid of 3x4 pixel cells.
+    :param p: Probability of a grid cell being set to 1 (not occluded).
```
brianhhu, line 24: Quick question- should we rename this parameter `p1` (here and below) to be consistent with RISE? I had some confusion whether the probability `p` corresponded to setting a grid to 0 or 1 (in some earlier discussions with Elim), and adding the 1 may help resolve some of this ambiguity.

brianhhu (commented):
Can we also add this new implementation to our docs [here]?

xaitk_saliency/impls/perturb_image/random_grid.py
schencej: I think that makes sense. I'll update and autosquash.

brianhhu:
LGTM, thanks!

PR #96, opened 2022-03-02
Add fill option to coco utils

schencej:
There was previously no way use a fill other than black when using `gen_coco_sal()` or `sal-on-coco-dets`. I added `fill` as an optional parameter that is passed through to `occlude_image_batch()`.

For now, it is assumed that a multi-channel fill value matches the number of channels of every image in the provided serialized detections, and similarly an image fill value matches the shape of each image.

`click` requires options to have explicit types, so for `sal-on-coco-dets` the fill is limited to 3 channel values.


xaitk_saliency/utils/bin/sal_on_coco_dets.py
```diff
@@ -25,6 +25,11 @@
 @click.argument('coco_file', type=click.Path(exists=True))
 @click.argument('output_dir', type=click.Path(exists=False))
 @click.argument('config_file', type=click.File(mode='r'))
+@click.option(
+    '--fill',
+    help='RGB values (0-255) of fill to use when occluding three dimensional images, will fail if used with 1D image',
```
brianhhu, line 30:
```suggestion
    help='RGB values (0-255) of fill to use when occluding images, will fail if used with 1-channel images',
```

Is there an elegant way to support passing either RGB values or a grayscale value? I'm thinking longer-term of potential use cases where detection might need to be run on grayscale imagery (e.g. from IR sensors). I guess more broadly, is this tool designed specifically for RGB images, or it supposed to be more general?

xaitk_saliency/utils/gen_coco_sal.py
```diff
@@ -55,6 +56,11 @@ def gen_coco_sal(
         :param img_perturber: Implementation of ``PerturbImage`` to use.
         :param sal_generator: Implementation of
             ``GenerateDetectorProposalSaliency`` to use.
+        :param fill: Optional fill for the occluded regions of each image as a
+            scalar value, a per-channel sequence or a shape-matched image. If
+            a multi-channel value is used, all the images in ``dets_dset`` must
+            have that number of channels. Similarly, if a shape-matched image
+            is used, it must match the shape of each image in ``dets_dset``.
```
brianhhu, line 63: Since we are passing this through to the `occlude_image_batch` function, should the `fill` parameter be synced to the docstring there? We mention alpha-blending (I think this was used for blurring), although I'm not 100% sure it is applicable here, e.g.

```suggestion
        :param fill: Optional fill for alpha-blending based on the input masks for
            the occluded regions as a scalar value, a per-channel sequence or a
            shape-matched image. If a multi-channel value is used, all the images
            in ``dets_dset`` must have that number of channels. Similarly, if a
            shape-matched image is used, it must match the shape of each image
            in ``dets_dset``.
```

xaitk_saliency/utils/bin/sal_on_coco_dets.py
```diff
@@ -59,6 +65,8 @@ def sal_on_coco_dets(
     :param config_file: Config file specifying the ``DetectImageObjects``,
         ``PerturbImage``, and ``GenerateDetectorProposalSaliency``
         implementations to use.
+    :param fill: RGB values for fill to use when occluding three-channel images. If your annotations
+        include one-channel images, this will fail.
     :param overlay_image: Overlay saliency maps on images with bounding boxes.
         RGB images are converted to grayscaleDefault is to output write out
         saliency maps by themselves.
```
brianhhu, line 72: Could you also check this param description- seems like we're missing a period maybe?

brianhhu (commented):
Overall, LGTM- please see my minor comments. I have a question regarding generality of this tool- do we envision it to only work with RGB images or potentially also grayscale imagery as well?

xaitk_saliency/utils/bin/sal_on_coco_dets.py
Purg: I chatted with @schencej earlier about using the click multiple value functionality which I think he is working on. With that you could provide something like `-f INT -f INT -f INT` instead of `--fill INT INT INT`, or alternatively just `-f INT` for a single-channel collection of images.

xaitk_saliency/utils/gen_coco_sal.py
brianhhu: Actually, thinking more about this, I think the tool can probably only support occlusion, i.e. the shape-matched fill value doesn't make sense here? Our CLI tool probably also doesn't support passing in images.

schencej: This description is in `gen_coco_sal`, the programmatic tool. If you are using this just as a function in a program you can pass an image.

brianhhu (approved):
LGTM, thanks! Feel free to ignore, but would it make sense to update our SerializedDetectionSaliency notebook to use a fill value? This may improve the quality of our saliency results (not 100% sure though).

Purg:
> LGTM, thanks! Feel free to ignore, but would it make sense to update our SerializedDetectionSaliency notebook to use a fill value? This may improve the quality of our saliency results (not 100% sure though).

I'm not opposed but I don't think we should block this PR with that.

schencej:
Updated the SerializedDetectionSaliency example. Saliency differences are also improved from the implementation changes to `gen_coco_sal` when only one prediction and label is provided.

Purg:
@brianhhu if the notebook changes look alright, I would like to get this in so I can make the release.

brianhhu:
LGTM- thanks everyone! This is turning out to be a really nice release.

PR #97, opened 2022-03-02
Add example comparing two models using saliency

schencej:


Purg:
Prelim review thoughts:
- notebook should be added to the CI list (or we should try at least), otherwise we should not somewhere why we do not include it in the list for CI (likely adding the "why" near the list in the CI config).
- "Download Detections and Image" section --> I know we did this in the centernet-only example, but I'm starting to feel like the approach of bringing in outside detections into this breaks the flow/compelling-factor of the example. It might be more compelling to run each detector on the image in deterministic modes, and then cherry-pick the specific detections (by ID, index position, etc.) that we want to highlight for comparison (the "we've done the exploration offline and distilled that here" part). What are your thoughts @schencej @brianhhu?
- "The id's for the annotations in each set are parallel." sentence: I don't understand what this means. If we use these precomputed detections files, we should also go into more detail on how they were generated and what they specifically contain, e.g. the top 10 most confident detections, or that we have a priori explored the and collated comparable detections on this image between these two detectors which we have distilled and serialized into these files. 
- "Note that the bounding boxes displayed overlap closely." sentence: we should add a "why" component to this, like "which is an effect of both models detecting the same object independently, which is a qualifier for our decision to highlight these specific detections."
- We might want to choose different colors than red and green here, as well as increase the thickness of drawn boxes. Red and green are colors naturally prevalent in this scene, so we should use colors that are not, e.g. cyan or magenta.
- We might want to make note that the `RandomGrid` choice follows in line with the `RISEGrid`, but more applicable to this use-case because of [insert reasons here].
- Clone Repository: "... allow us to easily use ..." --> "... allow us to use for this example ...". Probably want to add a warning or something that this is not good practice, but is for the same of this self-enclosed example.
- "... detections results in a common format." --> "... detections results in **_the_** common format, as defined by the interface."
- In the script-like cell where we're defining the yolo plugin, we should still follow flake8 style rules even though the tool does not know how to check notebook files, e.g. two newlines between module-level definitions, single new-line between intra-class definitions, etc. Ideally type annotations if only for documentation/readability, but that's not really a first-level priority.
- We might want to turn `gen_coco_sal(..., verbose=False)` for the saved notebook output. Verbose mode is a bit distracting for the notebook flow.
- "Compare Saliency" section -- I think the word description and the plotting are in conflict: text says yolo is the first row of the output, but the code says second row.

brianhhu:
Adding a few minor comments too:

1) I agree with Paul that it probably makes sense to generate a larger set of detections on the currently chosen input image, and then select a subset (e.g. interesting overlapping detections between the two models) for further explainability purposes. This removes some of the "hand-wavy" nature of how we get these detections from an external source.

2) Is there any way to suppress the [X/1200] and [X/400] verbose output from running the `gen_coco_sal` tool? I feel it adds a lot of rows to the output (which means more scrolling to do), but doesn't bring much value.

3) Currently `use_cuda = True` in the DetectImageObjects implementations- we'll probably want to shift computation to CPU if we want to be able to run the notebook in CI.

4) Why are only 400 perturbed images being generated for the CenterNet model?

5) I think adding some labels to the figure in the last cell to more clearly delineate which row corresponds to the YOLO model and which corresponds to the CenterNet model would be helpful for the reader, in case they just scroll through images.

Thanks!

schencej:
I agree that It makes more sense to run the detectors on the image first. Would switching to the new DRISE high-level interface make more sense here? It seems a bit weird to use `gen_coco_sal()` if we are not using serialized detections. @brianhhu @Purg

schencej:
> Why are only 400 perturbed images being generated for the CenterNet model?

`400` is the number of batches. It is using 1200 masks but the batch size for the CenterNet model is set to 3 so there are 400 batches. I will make this more clear.

brianhhu:
> > Why are only 400 perturbed images being generated for the CenterNet model?
> 
> `400` is the number of batches. It is using 1200 masks but the batch size for the CenterNet model is set to 3 so there are 400 batches. I will make this more clear.

OK, thanks for the clarification- this may be a moot point if we turn off the verbose output as this detail about how images are batched would be hidden from the user.

Purg:
> I agree that It makes more sense to run the detectors on the image first. Would switching to the new DRISE high-level interface make more sense here? It seems a bit weird to use `gen_coco_sal()` if we are not using serialized detections. @brianhhu @Purg

I'm with you on likely not using `gen_coco_sal` in the "do detections in notebook" case.

Purg:
Rebased on new master.

schencej:
The new example uses an interactive figure to select reference detections by clicking on them. I tried to keep the logic for this simple while still providing good functionality for the user. There are a couple faults:

- clicking where two predictions overlap selects both of them
- selecting a prediction that only one model has results in an uneven amount of saliency maps at the end

IMO keeping these "bugs" is fine for the purposes of this example and to keep the logic simple.

brianhhu:
I took a quick look at the notebook, overall the flow looks pretty good. A couple of questions:

1) Do we need the `tqdm` and `seaborn` dependencies? I don't see them being used in the notebook itself, but I didn't fully trace the YOLO code.

2) I saw that model inference time for the CenterNet model is about twice as long (1.5 s vs 0.75 s)- why is that exactly? Is this a function of num_workers or image size that can be controlled similar to the YOLO model?

3) Is there a potential case where the user selects a detection that only applies to a single model (e.g. tricycle in lower left)? Wouldn't this mess up the comparative saliency visualization?

examples/ModelComparisonWithSaliency.ipynb
```diff
@@ -0,0 +1,1843 @@
...
...hNTcCulbG6uoLGRh1Op42CbaFRr4sMPjtRRaFWRMbW0O9uYfXSRWEyN9cZ3aMhk7Wh56qojM3RcgmdrQbaZGRbTXj91KiKRTtniTkHKDN/iSIKgmfdcBgHr70GH/3oR0WSnEYTibAdr33ta/E///APhelnPBH3KsEC5w7JghPk8uvYAHE9X84x5b5JxkWkpoZZdHol88dZaYIP28xio+Fg3B7HntoudPwYxqUmlMubWJ3Q4St0kCU4sYVlpkmakaSGaW5xClu9PprrS7BUD1vryyiXitjY3MDI6CgK+QKa7T50IwcvTLBjcRGz83M4f/E8Lp05Bw85dKM8YFfl7JVzKhZnqpiarELVBzAsT2Slbj/BoBug16WMXUFAxi7yMWLbaB0/jebTZ6H4Pgb+AH7ZhzpuoWv2oGcTJIGPHmcmaSw3iGHRTMo0pUFGI6ReNxCpL+XlfG6algHbIuvopWc1jGXv8VkksUbgWESqOjBNoFrLo5Cz4PQcFAs1rK010O+mgJbPvsmZcWTKuoxk0Mxsc6spbr1kLsl8ZvWcuETTLIjz6NJElNEKjkKE4uKcynZV5AsmpufKMG1PwKnPOVKPM+8EypxvrYjDNl2GmQObRumYIhfnzKfTD9Fs9JAx89vMZgRd1WUmNfSlRSDPHnEbp5twkCoyGJPD2Vveiwh9zM6OY/eeBTz22BMC2hk5I4A/4Wx5Olfd6zkif6YZWCp7JvilmV2q5LkSQ8PmFVnUpWOXhzXvd6KQGL7HcAW+DSswPKzfhkUcvsRwBX5QVuDjf/s2wjsBUhwMYvFJ8wmaEUEtY3JqAbppieMnkgC9JgvrLeSLNiItBOeOCEoJUllME2SmgJNJE5TfEigRRybbDE+UBpTw67bnqvhQIsvDvwvJoFKSKIwCi74MbM73UeLGDrrD6IZAijSZPSUIo3RYI4ubvq6AVIJm/h3NQhFKniAlg1Acic5QYgOmlhdWNQhcYWJF+iZyO1Ukr73+QAw+aKz08le8QqI+KqUaLDuL+x/4NO779H3y/dPj86iOjUIzLZHo2oYuMj7Kh2MLMJnVqKp44NP3i6HM5Pgkrr/meuzfvQ+9ThOfuuvjyNsZVAsVRH6IRmMDCiVylREsrSxDNQ2EiAScDnxXADNZQrrlEnhR2inrR+dLK4O+xxVO3TcLxYzkHxKk9p0uRkaLCOMuBmRFOH
```
Purg, line 1646: Looks like the comment is on the wrong line.

Purg (commented):
I don't think I'm much of a fan of the interactive component in this specific context. It can be cool and allow for some flexibility of what you want to try out, but it also adds a level of complexity that requires us to stop and explain more things, and support more technical corner cases, that are orthogonal to the example and results we are trying to highlight.

Of specific distraction for me was the technical incompatibilities, e.g. this cannot be run in jupyter lab (which was the particular first go-to for our current audience), and the easy ability for me to pick bad examples when I did get it working. While picking bad examples is not without its value, we likely want that to happen in these examples with appropriate surrounding explanations as to what led to the "bad" of it. With this level of interactivity without a more complex support structure, the bad examples are without that critical context. I bet we *could* do that, technically speaking, but I think that's a much larger scope than intended for this example, as well as we start entering into "it's a tool" territory, which I also think we want to avoid here, if for no other reason than perception.

examples/ModelComparisonWithSaliency.ipynb
Purg, line 1620: (trying a second comment, not sure why the last one used the wrong line...)

The use of "in-line" doesn't make as much sense to me here (my brain goes to "in-line" code/compilation from C++ or other languages). Maybe:

```suggestion
    "This implementation is very similar to the `DRISEStack` algorithm, but allows us to define the granularity of our saliency maps in pixels whereas `DRISEStack` uses rigid square-grids that are deformed to cover an input image which results in less control in the shape of masked regions.\n",
```

Purg:
OK, well, I'm not sure why my notebook comments are showing up as referring to the wrong lines, so I'll just manually comment here and delete those other ones:

---
`Saliency Generation Object` section, sentence starting with "This is in-line with the `DRISEStack` algorithm, ..."

The use of "in-line" doesn't make as much sense to me here (my brain goes to "in-line" code/compilation from C++ or other languages). Maybe an alternative to this sentence could be:
"This implementation is very similar to the `DRISEStack` algorithm, but allows us to define the granularity of our saliency maps in pixels whereas `DRISEStack` uses rigid square-grids that are deformed to cover an input image which results in less control in the shape of masked regions."

---
```python
sal_generator = RandomGridStack(
    n=1200,  # number of masks
    s=(7,7),  # occlusion grid cell size in pixels
    p1=0.7,  # probability of not occluding cell
    threads=8,  # seed for replicable results
    seed=42
)
```

The comment on the `threads` line looks like it should be one down.

schencej:
Updated the example based on our discussion.

Also added a new commit to add the 'fill' parameter to the constructors of `PerturbationOcclusion` and the other implementations that use it. This allows a fill to be included in the configuration of these classes. @Purg and I had a discussion about this and decided just adding `fill` to the constructor is the best option for now.

brianhhu:
This LGTM- thanks @schencej for getting this done! I had a quick question about our environment setup cell:

```
!{sys.executable} -m pip install -q git+https://github.com/xaitk/xaitk-saliency.git
```

Do we have to install `xaitk-saliency` directly from source since some of these new features are not yet included in our PyPI package?

Purg:
> Do we have to install `xaitk-saliency` directly from source since some of these new features are not yet included in our PyPI package?

Yep. We just recently added the high-level interface and the `RandomGridStack` implementation used here in cell 10.

Purg (approved):
LGTM!

PR #99, opened 2022-03-08
Update algorithm table and README

brianhhu:
Minor updates to algorithm table in docs and README to help new users better understand supported saliency algorithms.

### DIFF
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -318,6 +318,7 @@ an algorithm, which is concerned only with the inputs and outputs.
 Additionally, the algorithm can be flexibly determined by the user; that is, the user is free to choose and configure
 the algorithm as needed for the problem domain.
 
+:ref:`custom text <Class: SlidingWindow>`
 
 The saliency algorithms can also be organized according to their respective tasks:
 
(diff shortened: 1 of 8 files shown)


brianhhu:
I was thinking a very similar thing today when editing the table, I will try and update that. Thanks!

Purg:
I *knew* there was a way to use inter-documentation references instead of hard-linking to the readthedocs website. This can be a "problem" in situations like when building your own offline documentation: built docs would then still make requests to the internet. I'm adding a commit that uses cross-referencing to refer to our local docs instead.

Purg (approved):
Received an LGTM from @brianhhu in an offsite chat.

docs/introduction.rst
```diff
@@ -318,6 +318,7 @@ an algorithm, which is concerned only with the inputs and outputs.
 Additionally, the algorithm can be flexibly determined by the user; that is, the user is free to choose and configure
 the algorithm as needed for the problem domain.
 
+:ref:`custom text <Class: SlidingWindow>`
```
brianhhu, line 321: @Purg Realized this snuck in on the PR (seems like a leftover artifact from early debugging tests). Let's try to remove this when we update the saliency algorithms table with the new DRISEStack implementation as part of Elim's PR.

PR #100, opened 2022-03-09
Add high-level interface and impls for blackbox detectors

schencej:


schencej:
I've realized that the we accept reference objectness scores as an input to `generate` which assumes that your model returns such a score. But the `DetectImageObjects` interface has no option to return an objectness score, so the perturbed detections never include it. Thoughts @Purg ?

xaitk_saliency/interfaces/gen_object_detector_blackbox_sal.py
```diff
@@ -0,0 +1,175 @@
...
+        class score should match the order returned by the input black-box
+        algorithm. The optional input objectness is expected to be a vector of
+        length `nDets` containing the objectness score for each reference
+        detection. If this is not provided, it is assumed that each detection
+        has a score of 1.
+
+        The output saliency map matrix should be (1) in the shape
+        `[nDets x H x W]` where `H` and `W` are the height and width
+        respectively of the input reference image, (2) floating-point typed,
+        and (3) composed of values in the `[-1, 1]` range.
```
Purg, line 54:
```suggestion
        and (3) composed of values in the `[-1, 1]` range.

        The `(0, 1]` range is intended to describe regions that are positively salient, 
        and the `[-1, 0)' range is intended to describe regions that are negatively salient.
        Positive values of the saliency heatmaps indicate regions that increase detection
        locality and class confidence scores, while negative values indicate regions that
        actively decrease detection locality and class confidence scores.
```

xaitk_saliency/interfaces/gen_object_detector_blackbox_sal.py
```diff
@@ -0,0 +1,175 @@
...
+            visual saliency maps for. This should be a matrix with shape
+            `[nDets x nClasses]`. The order of the scores should match that
+            returned by the input black-box detection algorithm.
+        :param blackbox: The black-box object detector to perform arbitrary
+            operations on in order to deduce visual saliency.
+        :param objectness: Optional objectness score for each reference
+            detection. This should be a vector of length `nDets`. If not
+            provided, it is assumed that each detection has a score of 1.
+
+        :raises ShapeMismatchError: The implementation result visual saliency
```
Purg, line 71: We should also include that ValueError exceptions may be raised.

xaitk_saliency/interfaces/gen_object_detector_blackbox_sal.py
```diff
@@ -0,0 +1,175 @@
...
+        output = self._generate(
+            ref_image,
+            bboxes,
+            scores,
+            blackbox,
+            objectness,
+        )
+
+        # Check that the saliency heatmaps' shape matches the reference image.
+        if output.shape[1:] != ref_image.shape[:2]:
```
Purg, line 112: We could also check that the quantity of output heatmaps matches the quantity of reference detections input.

xaitk_saliency/interfaces/gen_object_detector_blackbox_sal.py
```diff
@@ -0,0 +1,175 @@
...
+        scores: np.ndarray,
+        blackbox: DetectImageObjects,
+        objectness: Optional[np.ndarray] = None,
+    ) -> np.ndarray:
+        """
+        Internal method for implementing the generation logic.
+        This is invoked by the above `generate` method as a template method.
+
+        The doc-string for the `generate` method also applies here aside from
+        the `ShapeMismatchError` which is specific to `generate`.
```
Purg, line 154: We could also add that certain guarantees are being levied on the input parameters given the ValueError checks in `generate`.

tests/interfaces/test_gen_object_detector_blackbox_sal.py
```diff
@@ -0,0 +1,246 @@
...
+
+    # 2-channel image as just HxW should work
+    test_image = np.ones((256, 256), dtype=np.uint8)
+    GenerateObjectDetectorBlackboxSaliency.generate(
+        m_impl,
+        test_image,
+        test_bboxes,
+        test_scores,
+        m_detector,
+    )
```
Purg, line 34: It seems like we should be checking after each of these calls for something along the lines of `m_impl._generate.assert_called_with(...)`.

tests/interfaces/test_gen_object_detector_blackbox_sal.py
```diff
@@ -0,0 +1,246 @@
...
+        GenerateObjectDetectorBlackboxSaliency.generate(
+            m_impl,
+            test_image,
+            test_bboxes,
+            test_scores,
+            m_detector,
+        )
+
+
+def test_generate_checks_output_mismatch() -> None:
```
Purg, line 186: Aligning with the interface comment, we could add a congruent test that makes sure we check the quantity of heatmaps output against the number of detections input.

xaitk_saliency/impls/gen_object_detector_blackbox_sal/occlusion_based.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def get_config(self) -> Dict[str, Any]:
+        return {
+            "perturber": to_config_dict(self._perturber),
+            "generator": to_config_dict(self._generator),
+            "threads": self._threads,
+        }
+
+
+def _dets_to_formatted_mat(
```
Purg, line 115: There is a similar function to this in `xaitk_saliency/utils/gen_coco_sal.py`. Is there any deduplication that we can do with that, e.g. a common function in `xaitk_saliency/utils/`?

xaitk_saliency/impls/gen_object_detector_blackbox_sal/occlusion_based.py
```diff
@@ -0,0 +1,172 @@
...
+            [
+                bbox.min_vertex[0],
+                bbox.min_vertex[1],
+                bbox.max_vertex[0],
+                bbox.max_vertex[1]
+            ] for bbox in [det[0] for det in img_dets]
+        ])
+
+        score_dicts = [det[1] for det in img_dets]
+        scores_mat = np.array([list(d.values()) for d in score_dicts])
```
Purg, line 153: Relying on dictionary values iterable ordering seems a little dangerous/fragile to me. Maybe an input to this method could be a class label list to drive the scores extracted / order here?

tests/impls/gen_object_detector_blackbox_sal/test_drise.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def test_fill_prop(self) -> None:
+        """
+        Test that the `fill` property appropriately gets and sets the
+        underlying `PerturbationOcclusion` instance fill instance attribute.
+        """
+        inst = DRISEStack(5, 8, 0.5, seed=0)
+        assert inst.fill is None
+        inst.fill = 42
+        assert inst._po.fill == 42
```
Purg, line 103: Small detail but I would also add an `assert inst.fill == 42` to check property retrieval.

brianhhu:
Overall, LGTM- thanks @Purg for the detailed comments. Some minor suggestions:

1) We should probably update [the saliency algorithms table](https://github.com/XAITK/xaitk-saliency/blob/master/docs/introduction.rst#saliency-algorithms) in the intro to point to this new implementation (vs. just `DRISEScoring`).

2) We had a [note](https://github.com/XAITK/xaitk-saliency/blob/master/xaitk_saliency/impls/gen_detector_prop_sal/drise_scoring.py#L24-L27) in the `DRISEScoring` implementation about how to use DRISE with a single class prediction and confidence score. I think it might also make sense to either move or copy that docstring here to this implementation too.

3) We haven't updated the DRISE.ipynb notebook as part of this PR- would it make sense to switch to this implementation here and remove some of the current boilerplate code in that notebook?

4) Finally, now we have two potential DRISE-based tools, `gen_coco_sal` and `DRISEStack`. Can we make it somewhat more clear when one method is preferred over the other (e.g. if you have input detections, you should use `gen_coco_sal`)? Relatedly, there is no autodocs for the `gen_coco_sal` tool, which might be useful to help distinguish these two implementations.

xaitk_saliency/impls/gen_object_detector_blackbox_sal/occlusion_based.py
```diff
@@ -0,0 +1,172 @@
...
+        scores: np.ndarray,
+        blackbox: DetectImageObjects,
+        objectness: Optional[np.ndarray] = None,
+    ) -> np.ndarray:
+        ref_dets_mat = format_detection(bboxes, scores, objectness)
+
+        pert_masks = self._perturber(ref_image)
+
+        pert_dets = blackbox.detect_objects(
+            occlude_image_streaming(
```
Purg, line 67: Did we want to use the new threaded "batch" version here now instead? Or does that still use too much RAM with the quantity of perturbations used?

Purg:
> 3. We haven't updated the DRISE.ipynb notebook as part of this PR ...

That's not a bad idea. I believe an equivalent change was done when we updated the classifier black box interface.
We're pretty low on task parallelism right now, so this change being part of this PR seems fine to me.

> 4. Two tools

My vision in this regard is that the `gen_coco_sal` is for the CLI interface. I believe @schencej and I have chatted about this before, but I'm happy to talk about it if desired. The majority of the `gen_coco_sal` is just the `xaitk_saliency/impls/gen_object_detector_blackbox_sal/occlusion_based.py`'s `PerturbationOcclusion._generate()` method, but using kwcoco semantics for input detection extraction + formatting. We discussed last that with the "high-level" interface and implementations, *those* become the thing to use when programming, and the `gen_coco_sal` stuff is effectively "CLI only." Of course, given that, a chunk of, if not all of, the contents of that util could be shifted into the tool's script (`xaitk_saliency/utils/bin/sal_on_coco_dets.py`).

A ramification of that is that we will need to update the `examples/SerializedDetectionSaliency.ipynb` notebook to use the `GenerateObjectDetectorBlackboxSaliency` interface/implementation as it is taking the programmatic approach (as opposed to the CLI).

Purg:
> Yeah, maybe I wasn't very clear. I wasn't necessarily proposing transitioning `gen_coco_sal` to use the new implementation in this PR (even though there is some shared code). I was mainly thinking about adding additional documentation (maybe in ReadTheDocs) that would help a new user decide whether to use the CLI tool or the programmatic interface. I imagine there could potentially be some confusion since a user would have to dive into the docstrings of each function to understand the differences and use cases.

That's fair. Part of what I was trying to say above is that the `gen_coco_sal` function as it is could likely just "go away", as the remaining details of it are, I think, CLI specific and could be rolled right into the CLI tool. As a fallout of that, none of the library methods exposed would use kwcoco objects/containers, which I currently think is OK (but may be convinced otherwise).

xaitk_saliency/impls/gen_object_detector_blackbox_sal/occlusion_based.py
schencej: Added a class label input to do this. It grabs the labels from the first perturbed detection result it finds and uses those for the rest.

xaitk_saliency/impls/gen_object_detector_blackbox_sal/occlusion_based.py
schencej: I did some short tests of both and found that the difference in RAM usage was almost nothing, even with a high number of masks, so I switched to the batched version.

xaitk_saliency/impls/gen_object_detector_blackbox_sal/occlusion_based.py
schencej: The function in `gen_coco_sal` is a bit different, it uses the categories from the `kwcoco.CocoDataset` to maintain constant score ordering. I think it might make sense to convert both of these to a common function, but that would require changing the logic in `gen_coco_sal`. Since we are discussing getting rid of or changing `gen_coco_sal`, I think that can wait.

docs/implementations.rst
```diff
@@ -59,25 +59,39 @@ Class: SimilarityScoring
...
+--------------------------------
 .. autoclass:: xaitk_saliency.impls.gen_classifier_conf_sal.squared_difference_scoring.SquaredDifferenceScoring
   :members:
 
 ---------------------------------------------
 End-to-End Saliency Generation
 ---------------------------------------------
 
+Image Classification
+---------------------
```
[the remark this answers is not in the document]
schencej: I aligned all of the under-characters that I could find in our docs and added a new commit.

xaitk_saliency/impls/gen_object_detector_blackbox_sal/occlusion_based.py
```diff
@@ -0,0 +1,196 @@
...
+        ))
+
+        # use class labels of first non-empty detection from first image
+        labels = []
+        for img_dets in pert_dets:
+            for det in img_dets:
+                labels = list(det[1].keys())
+                break
+            if len(labels) > 0:
+                break
```
Purg, line 84: This likely OK to have out here, but since the result is only used for the `_dets_to_formatted_mat` function, this logic could be moved into that function. I also suspect you could lazy-populate this static list on your first detection encounter in there instead of a priori flattening all detections for all images.

brianhhu:
Thanks, overall this LGTM. Looks like there are minor differences in the final saliency map value ranges for the new high-level interface and the previous DRISE notebook (on master), although the rough shape of the maps look similar (maybe this is due to the detections being selected?) Also, there is a code smell for cognitive complexity, but I assume that's OK.

examples/DRISE.ipynb
```diff
@@ -503,20 +349,26 @@
...
     ")"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "4e092529",
+   "metadata": {},
+   "outputs": [],
+   "source": []
```
Purg, line 364: Empty cell that we can remove.

examples/DRISE.ipynb
```diff
@@ -503,20 +349,26 @@
...
   }
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "yolo",
```
Purg, line 369: I'm not sure what we can do about these sections changing every time, if anything. If you think of anything (besides manually not committing these specific changes per-modified-notebook file, which is a little manual) I'm all ears.

schencej: We should be able to create a filter that deletes the `kernelspec` entry of the metadata, which should make it default to Python3 kernel (accoring to this [issue](https://github.com/kynan/nbstripout/issues/141) that I found). I'm investigating this now.

Purg (approved):

PR #102, opened 2022-03-15
Added atari example notebook to the CI run list

Purg:


Purg:
Good news: It runs and it's not the slowest! I think this means we can include it in the CI list.

PR #103, opened 2022-03-17
Remove mistaken addition

Purg:
Accidental "test" line for sphinx functionality made it through in a
previous branch. This removes it.

There is no change note due to modifying content for which one already exists in this pending set.


Purg:
Aaaaand already in the HL interface branch, nvm.

PR #104, opened 2022-03-23
Change download links from Google Drive to data.kitware.com

brianhhu:
Swaps out download links for demo resources to data.kitware.com, to be compatible with the JAIC T&E sandbox environment.


Purg (approved):
LGTM

PR #105, opened 2022-03-23
Change coco utils to use high-level interface

schencej:
Highlights:
- `gen_coco_sal()` is gone and in its place is `parse_coco_dset()` which takes a `kwcoco.CocoDataset` object and creates matrices for use with the new high-level detector interface
- `sal-on-coco-dets` was updated to use this new utility function and high-level interface combination
- Updated testing of both modules
- I added new logic to mock failed import of the extra dependencies even when they are installed
- Updated the serialized detections example as well


schencej:
Using a fill with the CLI tool is now done by specifying it in the configuration for the `GenerateObjectDetectorBlackboxSliency` implementation. The fill currently isn't accounted for in the configuration suite of the existing implementations. I will address this in #97 since I changed the format of `drise.py` there.

schencej:
I believe the `continue` statements showing as not being covered by testing are actually covered. This seems to be related to some CPython optimization stuff, see [this issue](https://github.com/nedbat/coveragepy/issues/198#issuecomment-399705984) I found.

pyproject.toml
```diff
@@ -88,6 +88,7 @@ sal-on-coco-dets= "xaitk_saliency.utils.bin.sal_on_coco_dets:sal_on_coco_dets"
 "impls.gen_image_classifier_blackbox_sal.occlusion_based" = "xaitk_saliency.impls.gen_image_classifier_blackbox_sal.occlusion_based"
 "impls.gen_image_classifier_blackbox_sal.rise" = "xaitk_saliency.impls.gen_image_classifier_blackbox_sal.rise"
 "impls.gen_detector_prop_sal.drise_scoring" = "xaitk_saliency.impls.gen_detector_prop_sal.drise_scoring"
+"impls.gen_object_detector_blackbox_sal.drise" = "xaitk_saliency.impls.gen_object_detector_blackbox_sal.drise"
```
Purg, line 91: This happens to include the `occlusion_based.PerturbationOcclusion` implementation because the `_BaseDRISE` encapsulates it, but we should be explicit and include the former as an explicit export for completeness:
```suggestion
"impls.gen_object_detector_blackbox_sal.drise" = "xaitk_saliency.impls.gen_object_detector_blackbox_sal.drise"
"impls.gen_object_detector_blackbox_sal.occlusion_based" = "xaitk_saliency.impls.gen_object_detector_blackbox_sal.occlusion_based"
```

Purg (commented):
We can probably rename the `gen_coco_sal` module to something simpler now, like maybe just `coco`.

xaitk_saliency/utils/gen_coco_sal.py
```diff
@@ -1,285 +1,74 @@
...
-            cat_map = _map_score_dict_keys(score_dicts[0].keys(), cats)
-
-            # assume class ids are arranged between 0 and the maximum id
-            scores_mat = np.empty([0, max_cat_id+1])
-            objectness = np.ones(len(img_dets))
-
-            for det_i, score_dict in enumerate(score_dicts):
-                score_array = np.zeros(max_cat_id+1)
+                    # do not include any non-prediction annotations
+                    continue
```
Purg, line 63: Due to the same "continue isn't covered because CPython", we could also add the `# pragma: no cover` comment here.

tests/utils/test_gen_coco_sal.py
```diff
@@ -1,67 +1,87 @@
...
+
+        test_img_files = ['test_image1.png', 'test_image2.png']
+        test_imgs = [np.array(Image.open(os.path.join(DATA_DIR, f))) for f in test_img_files]
+
+        test_img_aids = [test_dset.gid_to_aids[gid] for gid in [1, 3]]
+
+        # construct expected bounding box and score matrices
+        test_bboxes = []
+        test_scores = []
+        for aids in test_img_aids:
```
Purg, line 63: This is more or less replicating the logic in the actual code being tested, so this isn't a great test I think. In this kind of situation we should generally manually construct the data structure as we would expect the result to be and inject as little logic as possible unless there really does need to be something dynamic about it. If this data-structure is very large, we may want to pickle a serialization of it to load here (I don't think that's the case right now? ).

xaitk_saliency/utils/bin/sal_on_coco_dets.py
```diff
@@ -45,7 +36,6 @@ def sal_on_coco_dets(
     coco_file: str,
     output_dir: str,
     config_file: TextIO,
-    fill: Optional[Tuple[int]],
     overlay_image: bool,
     generate_config_file: TextIO,
```
Purg, line 40: `generate_config_file` doesn't have a doc-string entry.

tests/utils/test_sal_on_coco_dets.py
```diff
@@ -36,8 +37,21 @@ def test_warning(self, tmpdir: py.path.local) -> None:
         runner = CliRunner()
 
         if is_usable:
-            with patch('xaitk_saliency.utils.bin.sal_on_coco_dets.is_usable', False):
```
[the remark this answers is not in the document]
schencej: Mocking `is_usable` to false doesn't test the section of code that checks if `kwcoco` is installed and then sets `is_usable` to false. I thought it would be more thorough to do it this way so that those lines are covered.

Purg:
@schencej With the function name change to `parse_coco_dset`, it looks like the failing notebook needs an update to use that new name.

Purg (approved):
LGTM. Just needs an autosquash and I think we're good to go. @brianhhu any last comments?

brianhhu (approved):
LGTM, thanks. I made a few very minor comments, none of them are really blocking this PR though.

brianhhu:
LGTM, thanks- feel free to merge!

PR #106, opened 2022-03-24
Update behvaior of SlidingWindow to be window-size agnostic

schencej:
The window placements are now determined independent of the window size.

All of the windows are shifted to center them and ensure that each edge of the image is covered evenly.

Now, given a set stride and image size, any window size will result in the same number of masks.


docs/release_notes/pending_release.rst
```diff
@@ -22,6 +22,12 @@ Documentation
   the generic ``PerturbationOcclusion``, and ``DRISEStack`` which uses
   ``RISEGrid`` and ``DRISEScoring`` together.
 
+Implementations
+
+* Updated behavior of the ``SlidingWindow`` ``PerturbImage`` implementation. For
+  a given stride, the number of masks generated is now agnostic of the window
```
brianhhu, line 28:
```suggestion
  a given stride, the number of masks generated is now agnostic to the window
```

brianhhu (commented):
We have several example notebooks that make use of the original `SlidingWindow` implementation (e.g. OcclusionSaliency.ipynb, SimilarityScoring.ipynb, etc.). I assume the new generated saliency maps will differ slightly with this change. Do we want to re-run and update those notebooks as part of this PR?

Purg:
> We have several example notebooks that make use of the original `SlidingWindow` implementation (e.g. OcclusionSaliency.ipynb, SimilarityScoring.ipynb, etc.). I assume the new generated saliency maps will differ slightly with this change. Do we want to re-run and update those notebooks as part of this PR?

If they are indeed different, then we probably ought to in order to be accurate. Good catch!

schencej:
Added a new commit to update the relevant examples.

brianhhu (commented):
@schencej, thanks for updating the notebooks! Overall, this LGTM- I think an added benefit is that the CI runtime of each notebook has now been reduced with the lower number of masks required, without affecting saliency map quality. A few minor comments:

1) I made a minor word suggestion change to the pending release notes entry ("agnostic of" -> "agnostic to").

2) With this PR, I see we've largely switched our notebook headers over from the `!pip install` to the `!{sys.executable} -m pip install` pattern. There are still three notebooks that use the old pattern (i.e. ATARI, SuperPixel, and Radial)- do we want to switch those for uniformity sake?

3) I made a note earlier that unlike [`RISEStack`](https://github.com/XAITK/xaitk-saliency/blob/master/xaitk_saliency/impls/gen_image_classifier_blackbox_sal/rise.py#L17-L19), our current [`SlidingWindowStack` docstring](https://github.com/XAITK/xaitk-saliency/blob/master/xaitk_saliency/impls/gen_image_classifier_blackbox_sal/slidingwindow.py#L17) doesn't contain a reference to the two sub-implementations, i.e. `SlidingWindow` and `OcclusionScoring`. This could be helpful for new users to discover the relevant functionality within `xaitk-saliency`, since in our algorithms table we point only to the `SlidingWindowStack` implementation.

I'm not sure this in scope for this PR, but just wanted to put it out there.

schencej:
Amended the first commit to include the pending release and `SlidingWindowStack` docstring edits.
Added a new commit to update the remaining examples to use `sys.executable`. This is an important edit IMO because using `!pip` does not install the packages to the virtual environment associated with the my kernel.

Purg (commented):
LGTM, but I'll let @brianhhu have a final say since he had the last set of comments.

brianhhu (approved):
LGTM, thanks! Since we're already updating notebooks in this PR, can we add the "Open in Colab" header to the model comparison notebook? I realized we forgot to do that in the original PR. This is not blocking, so if it's out-of-scope, I'm happy to defer to a new PR.

schencej:
Added the colab link. Should be good to go.

brianhhu:
LGTM, thanks- feel free to merge!

schencej:
Noticed that `use_cuda` was set to True in the model comparison example. Amended the last commit to set it to false as we have done in our other examples. Sorry for the last minute edit.

PR #107, opened 2022-03-30
Add image similarity high-level interface

schencej:
Highlights:

- Added interface for computing image similarity saliency maps: `GenerateImageSimilarityBlackboxSaliency`
- Added two implementations: `PerturbationOcclusion` (following pattern of other interfaces) and `SBSMStack`
- Updated the `SimilarityScoring` saliency generation impl to return a [HxW] heatmap instead of [1xHxW]. This is consistent with the description of the interface that it implements.
- Updated the similarity example notebook to use `SBSMStack`. I got rid of the application style used by the previous version because the new interface takes care of all the logic performed by the "app". The previous version also mentions that it shows the use of multiple API implementations, but it only uses one set of perturbation/scoring implementations. I got rid of these mentions.
- Adopted consistent nomenclature for the input images. Now the image that saliency is computed for is called the "query image" and the other is the "reference image". "Query" sounds like the image that we are interested in, in my head, so this naming scheme makes sense to me. Open to changing this.


tests/interfaces/test_gen_image_similarity_blackbox_sal.py
```diff
@@ -0,0 +1,142 @@
...
+    test_ref_image = np.empty((50, 50))
+    test_query_image = np.empty((256, 256, 7))
+
+    # mock _generate result with matching height and width to query image
+    m_impl._generate.return_value = np.empty((256, 256))
+
+    # mock image descriptor generator
+    m_desc_generator = mock.Mock(spec=ImageDescriptorGenerator)
+
+    GenerateImageSimilarityBlackboxSaliency.generate(
```
Purg, line 27: We should additionally test here that:
1) the _generate method was called with the expected inputs
1) the result returned here is the same as what the generator output

examples/SimilarityScoring.ipynb
```diff
@@ -471,42 +324,97 @@
...
+    "show_sal_map(test_image2, test_image2, sal_map)"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
   }
```
Purg, line 418: Remove empty cell.

schencej:
Rebased to new master and addressed comments in fixup commits.

xaitk_saliency/impls/gen_descriptor_sim_sal/similarity_scoring.py
```diff
@@ -94,7 +95,8 @@ def generate(
         # Ensure saliency map in range [-1, 1]
         sal = np.clip(sal, -1, 1)
 
-        return sal
+        # return just HxW components
+        return sal[0]
```
brianhhu, line 99: See comment above about whether or not we want to remove the first dimension.

xaitk_saliency/interfaces/gen_descriptor_sim_sal.py
```diff
@@ -53,15 +53,15 @@ def generate(
         decrease image similarity scores according to the model that generated
         input feature vectors.
 
-        :param ref_descr_1:
-            First image reference float feature-vector, shape `[nFeats]`
-        :param ref_descr_2:
-            Second image reference float feature-vector, shape `[nFeats]`
+        :param ref_descr:
+            Reference iamge float feature-vector, shape `[nFeats]`
```
brianhhu, line 57:
```suggestion
            Reference image float feature-vector, shape `[nFeats]`
```

xaitk_saliency/interfaces/gen_descriptor_sim_sal.py
```diff
@@ -53,15 +53,15 @@ def generate(
...
+        :param query_descr:
+            Query image reference float feature-vector, shape `[nFeats]`.
```
brianhhu, line 59:
```suggestion
            Query image float feature-vector, shape `[nFeats]`.
```

xaitk_saliency/impls/gen_descriptor_sim_sal/similarity_scoring.py
```diff
@@ -8,20 +8,21 @@
...
+    query image, as well as the masks of the query image perturbations (as would
+    be output from a `PerturbImage` implementation.
 
     The perturbation masks used by the following implementation are
     expected to be of type integer. Masks containing values of type
     float are rounded to the nearest value and binarized
     with value 1 replacing values greater than or equal to half of
     the maximum value in mask after rounding while 0 replaces the rest.
 
+    The resulting saliency maps are relative to the query image.
```
brianhhu, line 24: I think the wording here is a bit ambiguous, I suggest adding the same text you used below for the interface:

```suggestion
    The resulting saliency map is relative to the query image.
    As such, it denotes regions in the query image that make it more or less
    similar to the other image, called the reference image.
```

examples/SimilarityScoring.ipynb
```diff
@@ -471,41 +324,89 @@
...
-    "    test_image2_path, # The change\n",
-    "    test_image1_path, # The change\n",
-    "    simbbox,\n",
-    "    slid_algo,\n",
-    "    similarity_alg,\n",
-    "    fill=simbbox.blackbox_fill,\n",
+    "show_sal_map(test_image2, test_image1, sal_map)\n",
+    "\n",
+    "# a -> a\n",
+    "sal_map = sal_generator(\n",
```
brianhhu, line 396: The saliency map ranges here for the last two image similarity saliency pairs are slightly different than in master, e.g. [0.0397, 1] vs. [0.0607, 1] and [0.044, 1] vs. [0.079, 1]. Do we know why this is? The shape of the saliency maps look generally consistent though.

brianhhu (commented):
Overall, LGTM. I made a few minor suggestions, and wanted to clarify whether we do actually want to remove the first dimension of the similarity-based saliency maps, as it makes it less consistent with our other saliency map outputs, which are generally of the form [N x H x W].

examples/SimilarityScoring.ipynb
schencej: I believe this is due to the updated sliding window implementation.

PR #110, opened 2022-04-08
Vectorize image similarity interfaces

schencej:
Similarity interfaces vectorized per out discussion.

One point of contention:
- Currently the query images input for the high-level interface is an `Iterable`. If we switched to `Sequence` we could check that the number of output heatmaps matches the number of input images


xaitk_saliency/impls/gen_descriptor_sim_sal/similarity_scoring.py
```diff
@@ -10,20 +10,14 @@ class SimilarityScoring (GenerateDescriptorSimilaritySaliency):
     """
     This saliency implementation transforms proximity in feature space into
     saliency heatmaps.
-    This should require a sequence of feature vectors of the query and reference
-    image, a number of feature vectors as predicted on perturbed versions of the
-    query image, as well as the masks of the query image perturbations (as would
-    be output from a `PerturbImage` implementation.
+    This should require feature vectors for the reference image, for each query
+    image, and for perturbed versions of the reference image, as well as the
+    masks of the query image perturbations (as would be output from a
```
Purg, line 15:
```suggestion
    image, perturbed versions of the reference image, the
    masks of the reference image perturbations (as would be output from a
```

xaitk_saliency/interfaces/gen_descriptor_sim_sal.py
```diff
@@ -10,35 +10,36 @@ class GenerateDescriptorSimilaritySaliency (Plugfigurable):
...
         perturbed_descrs: np.ndarray,
         perturbed_masks: np.ndarray,
     ) -> np.ndarray:
         """
-        Generate a visual saliency heatmap matrix given the black-box
-        descriptor generation output on two reference images, the same
-        descriptor output on perturbed images and the masks of the visual
+        Generate a matrix of visual saliency heatmaps given the black-box
+        descriptor generation output on a reference image, several query images,
+        perturbed version of the reference image and the masks of the visual
```
Purg, line 35:
```suggestion
        perturbed versions of the reference image and the masks of the visual
```

xaitk_saliency/interfaces/gen_descriptor_sim_sal.py
```diff
@@ -55,10 +56,10 @@ def generate(
 
         :param ref_descr:
             Reference image float feature-vector, shape `[nFeats]`
-        :param query_descr:
-            Query image float feature-vector, shape `[nFeats]`.
+        :param query_descrs:
+            Query images float feature-vectors, shape `[nQueryImgs x nFeats]`.
```
Purg, line 60: The double-plural sounds odd to me, but I'm happy for someone for fact-check my "grammar feel".
```suggestion
            Query image float feature-vectors, shape `[nQueryImgs x nFeats]`.
```

tests/interfaces/test_gen_image_similarity_blackbox_sal.py
```diff
@@ -16,10 +16,10 @@ def test_generate_checks_success() -> None:
 
     # test images
     test_ref_image = np.empty((50, 50))
-    test_query_image = np.empty((256, 256, 7))
+    test_query_images = [np.empty((256, 256, 7))]*3
```
Purg, line 19:
```suggestion
    test_query_images = [np.empty((256, 256, 7))] * 3
```

tests/interfaces/test_gen_image_similarity_blackbox_sal.py
```diff
@@ -43,76 +43,48 @@ def test_generate_checks_success() -> None:
...
+    Test that the input reference image conforms to our assumption.
     """
     m_impl = mock.Mock(spec=GenerateImageSimilarityBlackboxSaliency)
     m_desc_generator = mock.Mock(spec=ImageDescriptorGenerator)
 
-    # bad ref image
+    # too few dimensions
     test_ref_image = np.empty((256,))
-    test_query_image = np.empty((256, 400, 3))
+    test_query_images = [np.empty((256, 400, 3))]*2
```
Purg, line 53:
```suggestion
    test_query_images = [np.empty((256, 400, 3))] * 2
```

tests/interfaces/test_gen_image_similarity_blackbox_sal.py
```diff
@@ -43,76 +43,48 @@ def test_generate_checks_success() -> None:
...
     """
     Test that the `_generate` output shape is appropriately checked against that
-    of the query image.
+    of the reference image.
     """
     m_impl = mock.Mock(spec=GenerateImageSimilarityBlackboxSaliency)
 
     test_ref_image = np.empty((256, 123))
-    test_query_image = np.empty((100, 200, 3))
+    test_query_images = [np.empty((100, 200, 3))]*2
```
Purg, line 87:
```suggestion
    test_query_images = [np.empty((100, 200, 3))] * 2
```

xaitk_saliency/impls/gen_descriptor_sim_sal/similarity_scoring.py
```diff
@@ -53,35 +47,37 @@ def __init__(
...
-        if len(ref_descr) != len(query_descr):
-            raise ValueError("Length of feature vector between "
-                             "two images do not match.")
+        if len(ref_descr) != query_descrs.shape[1]:
+            raise ValueError("Size of feature vectors between reference and "
+                             "query images do not match.")
 
-        # Computing original proximity between image1 and image2 feature vectors.
+        # Computing original proximity between reference image and each query
+        # image feature vector.
```
brianhhu, line 63:
```suggestion
        # Computing original proximity between reference image feature vector
        # and each query image feature vector.
```

xaitk_saliency/impls/gen_descriptor_sim_sal/similarity_scoring.py
brianhhu: I think you need to add an `and` in the suggested sentence above.

xaitk_saliency/interfaces/gen_descriptor_sim_sal.py
```diff
@@ -10,35 +10,36 @@ class GenerateDescriptorSimilaritySaliency (Plugfigurable):
...
     saliency heatmaps.
 
-    This transformation requires two reference images, translated into
-    feature-vectors via some black-box means, between which we are trying to
-    discern the feature-space saliency.
-    This also requires the feature-vectors for perturbed images as well as the
-    masks of the perturbations as would be output from a
-    :class:`xaitk_saliency.interfaces.perturb_image.PerturbImage`
+    This transformation requires a reference image, and a number of query images
+    all translated into feature-vectors via some black-box means.
```
brianhhu, line 14: I noticed that we are sometimes inconsistent in whether we use "feature vector(s)" or "feature-vector(s)". I think all instances prior to this instance use the un-hyphenated version.

xaitk_saliency/interfaces/gen_image_similarity_blackbox_sal.py
```diff
@@ -25,57 +26,56 @@ class GenerateImageSimilarityBlackboxSaliency (Plugfigurable):
...
     ) -> np.ndarray:
         """
-        Generates visual saliency map based on the similarity of two images
-        determined by the output of the blackbox feature vector generator.
+        Generates visual saliency maps based on the similarity of the reference
+        image to each query image determined by the output of the blackbox
+        feature vector generator.
 
-        The input reference and query image are expected to be matrices in
+        The input reference image is expected to a matrix in
```
brianhhu, line 37:
```suggestion
        The input reference image is expected to be a matrix in
```

brianhhu (commented):
Overall, LGTM- thanks for the update! Added a few minor suggestions for typo and grammar fixes.

xaitk_saliency/impls/gen_descriptor_sim_sal/similarity_scoring.py
schencej: IMO, getting rid of the "and for" here makes it seem like the perturbed images themselves are required, not features vectors for those images. Similarly I think the "as well as" emphasizes that the required perturbation masks are not feature vectors.

schencej:
@brianhhu @Purg Did you see my initial comment above? I had an implementation question about the type of our query images input.

Purg:
> Did you see my initial comment above?

I did! I thought I wrote a comment related to that, but I see that I must not have (or missed pressing the "Comment" button).

Changing the requirement to be a `Sequence` would be the simplest way to allow that check at the interface level.

Alternatively, we can still acquire such a count without requiring an explicit sequence by creating a wrapper iterator around it:
```python
def generate(...):
    ...
    num_query = 0
    def wrap():
        nonlocal num_query
        for q in query_images:
            yield q
            num_query += 1

    # the same "get actual output" line as before, but changing the iterable input to the wrapper iterator
    output = self._generate(ref_image, wrap(), blackbox)

    # num_query should now be the number of query images
    assert output.shape[0] == num_query, f"Output heatmaps didn't match input images: {output.shape[0]} != {num_query}"

    ...
```

I suppose technically the iterator count be copied into separate processes (e.g. torch dataloader with iterable dataset), and there is a solution to that as well using `multiprocessing.Manager()`, but I an open to feedback from @brianhhu + others on if that complexity is OK for this point in time. I'm sure the following could be improved because the space of corner cases is a bit larger.
```python
def generate(...):
    ...
    with multiprocessing.Manager() as manager:
        num_query = manager.Value(int, 0)
        idx_obs = manager.Value(set, set())
        lock = manager.Lock()
        def wrap():
            nonlocal lock, count, idx_obs
            for i, q in enumerate(query_images):
                yield q
                with lock:
                    if i not in idx_obs.value:
                        num_query.value += 1
                        idx_obs.value.add(i)

        output = self._generate(ref_image, wrap(), blackbox)

        # num_query should now be the number of query images
        assert output.shape[0] == num_query.value, \
            f"Output heatmaps didn't match input images: {output.shape[0]} != {num_query.value}"

    ...
```

brianhhu:
I have no strong opinion on this matter, but what are the drawbacks of just making it a Sequence? At first glance, it looks like a lot of added complexity for that check if we decide to still use iterators. In the end, I'm open to whatever we think is best- thanks!

Purg:
> but what are the drawbacks of just making it a Sequence?

The only obvious drawback is that we would not be able to operate on an unsized container, e.g. a stream, but considering that use-case in this context is likely over-design. I think your comment on complexity is valid. We can change to the simpler `Sequence` type for input for now.

xaitk_saliency/impls/gen_descriptor_sim_sal/similarity_scoring.py
```diff
@@ -98,7 +94,7 @@ def generate(
         sal = np.clip(sal, -1, 1)
 
         # return just HxW components
```
brianhhu, line 96: This comment is extraneous given we reverted back to returning NxHxW saliency maps.

brianhhu (approved):
LGTM, thanks!

schencej:
Addressed the last comment and autosquashed. Should be good to go.

Purg (approved):

PR #115, opened 2022-06-08
Revise re-use of smqtk-core publish workflow

Purg:
It was the case that the workflow had a parent-org match check. The
workflow from smqtk-core would check for the "Kitware" org which is not
valid in this case. It was also the case that the shared workflow was
using an incorrect base version of python to build the package in (3.6
vs. 3.7).

This change creates a new version of this workflow with the appropriate
edits that bring into line with the XAITK org and other repository
specifics.

### DIFF
--- a/.github/workflows/ci-example-notebooks.yml
+++ b/.github/workflows/ci-example-notebooks.yml
@@ -6,11 +6,12 @@ on:
   push:
     branches:
       - master
-      - release*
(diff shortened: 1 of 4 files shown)


Purg:
@brianhhu PTAL

brianhhu (approved):
LGTM, thanks!

PR #124, opened 2022-09-28
Fix sphinx versions

Purg:
Building the sphinx documentation became broken with the pinned versions installed via poetry at some point in the recent past. This updates the versions of the sphinx-related dependencies (and smqtk-core for some reason, but that's OK) to the latest, which also brings us into parity with what RTD installs during it's build process (a.k.a. the latest versions).


brianhhu (approved):
LGTM, thanks!

PR #129, opened 2022-12-13
Update codecov action config is CI workflow

Purg:


Purg:
There seems to be a github branch acquisition hiccup that I don't understand affecting ghostflow and RTD checks. I didn't see these yesterday when doing a bunch of CI with smqtk repos yesterday so I'm trying to un-confuse myself.

Purg:
The issue seems to have something to do with this PR. The offending command from at least RTD's perspective *works* for different PRs, just not this one. I'm going to cycle a new PR to see if there was a specific hiccup with this PR, or if just new PRs since yesterday are being affected.

PR #131, opened 2022-12-13
More open abstract requirements

Purg:


brianhhu (commented):
Generally looks good to me. Just to check my understanding:

- What is the correspondence now between the `pyproject.toml`/`poetry.lock` files in `xaitk-saliency` and `xaitk-cdao`? They both seem to have some bifurcating dependencies based on Python version.

- Are the added `#type: ignore` comments a result of the new codecov action version?

- The updated superpixel saliency notebook seems to have to do with changing the input image type to ndarray- was that due to a skimage API change?

Thanks!

Purg:
> What is the correspondence now between the pyproject.toml/poetry.lock files in xaitk-saliency and xaitk-cdao? They both seem to have some bifurcating dependencies based on Python version.

The `poetry.lock` files are specific to one repo and only for those that have a use-case that requires checking out the repo (developers/CI/etc.) for creating reproducible environments. The `pyproject.toml` file is for package configuration and abstract dependency configuration (including developer specific deps that also only have effect when manipulating locally with poetry). The main abstract deps in the file are what are encoded effectively into the package when built and distributed (i.e. to PYPI).

Effectively, xaitk-saliency doesn't know nor should care about xaitk-cdao. On the other hand, xaitk-cdao, as it has xaitk-saliency as an abstract dependency, will inherit the same requirements. Currently xaitk-cdao has nearly no main abstract deps, nor is this branch's modifications going to be present in a released package just yet, so it will not need any modification to make use of xaitk-saliency. There is "duplication" in the dev dependencies because there is xaitk-saliency does not expose any dev dependencies (its only a poetry concept), so the duplication is just a "consistency" thing given the intention to support the same CI/documentation support (flake8/mypy/pytest/sphinx/etc.).

Purg:
> Are the added #type: ignore comments a result of the new codecov action version?

Are you specifically referring to the `# type: ignore[empty-body]`?

Either way, these additions were due to the updated version of mypy that is being pinned here, which has additional checks that have been added to it since the previous version. These ignores were added just support structures only used in unittests, since those structures are exceptional and don't follow "production practices" on purpose (i.e. not implementing a non-abstract method on purpose because it's a test stub).

Purg:
> The updated superpixel saliency notebook seems to have to do with changing the input image type to ndarray- was that due to a skimage API change?

I expect so. The offending method, `mark_boundaries`, "says" that it can accept a PIL images but it then immediately calls `image.dtype`, which is not  a valid property on PIL images, so... I just converted it to a numpy array immediately, which also jives with the rest of what is happening in there.

brianhhu:
> > What is the correspondence now between the pyproject.toml/poetry.lock files in xaitk-saliency and xaitk-cdao? They both seem to have some bifurcating dependencies based on Python version.
> 
> The `poetry.lock` files are specific to one repo and only for those that have a use-case that requires checking out the repo (developers/CI/etc.) for creating reproducible environments. The `pyproject.toml` file is for package configuration and abstract dependency configuration (including developer specific deps that also only have effect when manipulating locally with poetry). The main abstract deps in the file are what are encoded effectively into the package when built and distributed (i.e. to PYPI).
> 
> Effectively, xaitk-saliency doesn't know nor should care about xaitk-cdao. On the other hand, xaitk-cdao, as it has xaitk-saliency as an abstract dependency, will inherit the same requirements. Currently xaitk-cdao has nearly no main abstract deps, nor is this branch's modifications going to be present in a released package just yet, so it will not need any modification to make use of xaitk-saliency. There is "duplication" in the dev dependencies because there is xaitk-saliency does not expose any dev dependencies (its only a poetry concept), so the duplication is just a "consistency" thing given the intention to support the same CI/documentation support (flake8/mypy/pytest/sphinx/etc.).

I was mostly thinking about the `xaitk-saliency` abstract dependency in `xaitk-cdao`, which you helped explain- thanks!

brianhhu:
> > The updated superpixel saliency notebook seems to have to do with changing the input image type to ndarray- was that due to a skimage API change?
> 
> I expect so. The offending method, `mark_boundaries`, "says" that it can accept a PIL images but it then immediately calls `image.dtype`, which is not a valid property on PIL images, so... I just converted it to a numpy array immediately, which also jives with the rest of what is happening in there.

Cool, thanks!

brianhhu:
> > Are the added #type: ignore comments a result of the new codecov action version?
> 
> Are you specifically referring to the `# type: ignore[empty-body]`?
> 
> Either way, these additions were due to the updated version of mypy that is being pinned here, which has additional checks that have been added to it since the previous version. These ignores were added just support structures only used in unittests, since those structures are exceptional and don't follow "production practices" on purpose (i.e. not implementing a non-abstract method on purpose because it's a test stub).

Yes, I was seeing those added in-line comments- thanks for clarifying.

brianhhu (approved):
Thanks for clarifying my questions- LGTM!

PR #133, opened 2023-01-05
Small formatting update to benchmark util to make easier to read

Purg:
Since I was looking at this again, I thought it could be made more easy to read.


brianhhu:
@Purg and I decided offline to create a separate PR to address the SonarCloud failure, which is not necessary to fix at this time before merging. Thanks~

PR #134, opened 2023-01-05
Lowered required version after 3.8 hinge

Purg:
Revision to abstract specifications to be different yet low enough to be open to compatibility with other package requirements.

We need to have a different version specifier for after the hinge, so we only change the ">=" to ">". The versions should follow the minimum required for the python applicable python versions.


Purg:
Don't need to update the release notes, this is updating a change already noted.

bjrichardwebster:
This PR is getting superseded by #136 since increasing the minimum supported python version to 3.8 removes the need for these hinges.

PR #136, opened 2023-06-05
Update to Python 3.8 minimum.

bjrichardwebster:


bjrichardwebster:
Although there aren't any unittests for using the notebooks on py3.11, I also unpeggd torch and torchvision versions in the notebooks to support py3.11. 

One of the notebooks (examples/VIAME_OcclusionSaliency.ipynb) can't run with py3.11+ because of a zip that is downloaded from data.kitware.com that contains a python script with code baked in. That code requires an attribute from torchvision.models.resnet that was removed in 0.13.0+. However, no version torchvision<0.13.0 supports py3.11+. 

Since this notebook isn't part of the package when released, there isn't a critical need to have py3.11+ supported. I put an assert in the notebook (with an explanation) instead to trigger if using 3.11+ so a user of the notebook would know how to fix the issue on failure rather than digging into the downloaded zip that contains the script to find the error.

pyproject.toml
```diff
@@ -25,38 +25,34 @@ classifiers = [
...
+scipy = ">=1.8.1"
 click = ">=8.0.3"
+setuptools = "*"
 # Optionals for "example" extra
 jupyter = { version = ">=1.0.0", optional = true }
 matplotlib = { version=">=3.4.1", optional = true }
 papermill = { version = ">=2.3.3", optional = true }
-torch = { version = ">=1.9.0", optional = true }
-torchvision = { version = ">=0.10.0", optional = true }
+torch = {version = ">=1.13.1,!=2.0.1", optional = true}
```
Purg, line 54: To note, the dependabot warning that motivated this is more specifically pertaining to our *concrete* versioning listed in the `poetry.lock` file, not here. We discussed at an earlier time that we may encounter incompatibilities with other packages because we claim to not work with versions below 1.13.1, but we actually can. The "not recommended" part (dependabot) is external to if someone does. E.g. We could keep the lower bound on the abstract requirement at `1.9.0` while maintaining our use of >=1.13.1 in our concrete requirement like it currently is (set to `2.0.0`).

PR #141, opened 2024-01-30
Fix sonarcloud analysis issues

Purg:
Fixing now-failing SonarCloud analysis items.


brianhhu (approved):
LGTM, thanks! The majority of changes seem to be due to how we originally used `np.random`, which have now been updated in this PR.

PR #147, opened 2024-05-31
Fix numpy version issue

stephencrowell:


pyproject.toml
bjrichardwebster, line 1: Please change `numpy` versions to the following code-block and update `poetry.lock` to reflect the version change.

```
# Python capped to <3.13 because of:
# https://stackoverflow.com/questions/77431252/why-doesnt-poetry-select-the-good-version-of-my-dependency
numpy = [
    {version = ">=1.22,<1.25", python = "~3.8"}, # CVE-2021-34141
    {version = ">=1.22,^1.25", python = ">=3.9,<3.13" } # CVE-2021-34141
]
```

bjrichardwebster (approved):

PR #160, opened 2024-11-12
Add check for no detections

stephencrowell:
Checks to see if `_dets_to_formatted_mat` returns no detections and returns early. Also added a check to see if any saliency maps were generated in `GenerateObjectDetectorBlackboxSaliency`.

### DIFF
--- a/tests/impls/gen_object_detector_blackbox_sal/test_occlusion_based.py
+++ b/tests/impls/gen_object_detector_blackbox_sal/test_occlusion_based.py
@@ -12,6 +12,10 @@ from xaitk_saliency.impls.gen_object_detector_blackbox_sal.occlusion_based impor
 from xaitk_saliency.utils.masking import occlude_image_batch
 
 
+def _perturb(ref_image: np.ndarray) -> np.ndarray:
+    return np.ones((6, *ref_image.shape[:2]), dtype=bool)
+
+
 class TestPerturbationOcclusion:
 
     def teardown(self) -> None:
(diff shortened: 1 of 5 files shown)


tests/impls/gen_object_detector_blackbox_sal/test_occlusion_based.py
```diff
@@ -14,6 +14,31 @@
...
+
+        def generate(
+            self,
+            ref_dets: np.ndarray,
+            pert_dets: np.ndarray,
+            pert_masks: np.ndarray
+        ) -> np.ndarray:
+            return np.zeros((ref_dets.shape[0], *pert_masks.shape[1:]), dtype=np.float16)
+
+        get_config = None  # type: ignore
```
eveenhuis, line 40: Looks like these were here before, but now that we're adding even more trivial impls, does it make sense to use mocking instead? (Applies to other test impls too)

tests/impls/gen_object_detector_blackbox_sal/test_occlusion_based.py
```diff
@@ -160,3 +160,42 @@ def detect_objects(
...
             assert m_kwargs['fill'] == test_fill
+
+    def test_empty_detections(self) -> None:
+        """
+        Test invoking _generate() with empty detections.
+        """
+
+        class StubDetector (DetectImageObjects):
+            """
+            Stub object detector that returns known detections.
```
eveenhuis, line 171: Nit: this one doesn't return dets

xaitk_saliency/interfaces/gen_object_detector_blackbox_sal.py
```diff
@@ -144,6 +147,10 @@ def generate(
             objectness,
         )
 
+        if len(output) == 0:
+            logging.info("No detections found for image. Check DetectImageObjects and saliency configuation")
```
eveenhuis, line 151: Do we think this qualifies as a warning (more than info)?

stephencrowell: When I talked with @bjrichardwebster about the issue, we both agreed we did not want to clutter the console with the logging info. I thought that (_by default_) warning would be output to the console. If that is wrong I have no issues with changing this.

eveenhuis: Okay, with that, lets keep this as `info` then. Most loggers are set to WARN and above

tests/impls/gen_object_detector_blackbox_sal/test_occlusion_based.py
```diff
@@ -143,13 +127,18 @@ def detect_objects(
...
-            inst = PerturbationOcclusion(test_pi, test_gen)
-            inst.fill = test_fill
-            test_result = inst._generate(
+            m_impl = mock.Mock(spec=PerturbationOcclusion)
+            m_impl._perturber = _perturb
+            m_impl._generator = _generate
+            m_impl.fill = test_fill
+            m_impl._threads = 0
+
+            test_result = PerturbationOcclusion._generate(
```
eveenhuis, line 136: This seems weird to me. We're mocking _generate, but also testing it? Why aren't we calling this on `m_impl` directly? I don't think mocking should prevent that

tests/impls/gen_object_detector_blackbox_sal/test_occlusion_based.py
```diff
@@ -160,3 +149,39 @@ def detect_objects(
...
+        m_impl.fill = None
+        m_impl._threads = 0
+
+        test_result = PerturbationOcclusion._generate(
+            m_impl,
+            test_image,
+            test_bboxes,
+            test_scores,
+            m_detector,
+        )
```
eveenhuis, line 185: Same comment

tests/interfaces/test_gen_object_detector_blackbox_sal.py
```diff
@@ -301,3 +301,38 @@ def test_call_alias() -> None:
...
+    expected_return = np.array([])
+    m_impl._generate.return_value = expected_return
+
+    test_ret = GenerateObjectDetectorBlackboxSaliency.generate(
+        m_impl,
+        test_image,
+        test_bboxes,
+        test_scores,
+        m_detector,
+    )
```
eveenhuis, line 329: Same comment

eveenhuis:
LGTM, but CI is failing

PR #166, opened 2024-11-26
Updates from GitLab

eveenhuis:


eveenhuis:
@bjrichardwebster CI is still running as I write this, but ghostflow has already failed. If CI fails on GitHub (after passing on GitLab) what do you want the resolution strategy to be? Hopefully it doesn't happen *too* often, but clearly it's not impossible

bjrichardwebster:
> @bjrichardwebster CI is still running as I write this, but ghostflow has already failed. If CI fails on GitHub (after passing on GitLab) what do you want the resolution strategy to be? Hopefully it doesn't happen _too_ often, but clearly it's not impossible

When I was testing this out, I actually pushed to a PR to test it out actually, so I didn't catch this. What I thought about though in general is that there are required tests on github that we don't/can't have on gitlab, one of which is ghostflow. I believe that all of these will be very small fixes that the maintainers should be able to fix quickly. This ghostflow failure is an example where it is a bunch of trailing white space. That said, we should look at why all that exists and proprogate it back into devel-jatic when appropriate.

eveenhuis:
> > @bjrichardwebster CI is still running as I write this, but ghostflow has already failed. If CI fails on GitHub (after passing on GitLab) what do you want the resolution strategy to be? Hopefully it doesn't happen _too_ often, but clearly it's not impossible
> 
> When I was testing this out, I actually pushed to a PR to test it out actually, so I didn't catch this. What I thought about though in general is that there are required tests on github that we don't/can't have on gitlab, one of which is ghostflow. I believe that all of these will be very small fixes that the maintainers should be able to fix quickly. This ghostflow failure is an example where it is a bunch of trailing white space. That said, we should look at why all that exists and proprogate it back into devel-jatic when appropriate.

Should the failures be fixed directly in this PR or addressed on gitlab first?

bjrichardwebster:
> > > @bjrichardwebster CI is still running as I write this, but ghostflow has already failed. If CI fails on GitHub (after passing on GitLab) what do you want the resolution strategy to be? Hopefully it doesn't happen _too_ often, but clearly it's not impossible
> > 
> > 
> > When I was testing this out, I actually pushed to a PR to test it out actually, so I didn't catch this. What I thought about though in general is that there are required tests on github that we don't/can't have on gitlab, one of which is ghostflow. I believe that all of these will be very small fixes that the maintainers should be able to fix quickly. This ghostflow failure is an example where it is a bunch of trailing white space. That said, we should look at why all that exists and proprogate it back into devel-jatic when appropriate.
> 
> Should the failures be fixed directly in this PR or addressed on gitlab first?

For any of the GitHub-only failures, I think it is okay to just fix them here as a maintainer. We can update the doc to reflect this as well

PR #171, opened 2025-01-15
MC-RISE parallel_map bug fix

eveenhuis:


eveenhuis:
MNIST notebook failing due to cyber attack on openml, see [here](https://github.com/orgs/openml/discussions/20) for details

PR #183, opened 2026-02-10
Pin upper bound for setuptools to '<82.0.0'

barry-ravichandran:
@bjrichardwebster This PR sets the upper bound for `setuptools`. This should help resolve the issues our package is having while installing within the RI.


bjrichardwebster:
Replaced by https://github.com/XAITK/xaitk-saliency/pull/184

PR #187, opened 2026-09-10
Development from GitLab

stephencrowell:


bjrichardwebster:
@stephencrowell can we add something like this gatekeeper job to the workflow so we don't have to keep updating the required jobs?

https://github.com/orgs/community/discussions/4324#discussioncomment-3477871

.github/workflows/.github-test.yml
bjrichardwebster: Does caching no longer work because we're using tox?

.github/workflows/.github-test.yml
```diff
@@ -79,9 +108,14 @@ jobs:
           files: coverage.xml
           flags: unittests
           skip_validation: true
+
   notebooks:
+    name: notebook (${{ matrix.notebook-file }})
+    if: github.event_name == 'workflow_dispatch' && inputs.run_notebooks
```
bjrichardwebster: Does this ever get run? `workflow_dispatch` doesn't have any inputs. In the past we haven't done nightly builds on github and instead just ran all of the available notebooks. We just don't require them to pass.

.github/workflows/.github-test.yml
```diff
@@ -93,61 +127,20 @@ jobs:
             "SimilarityScoring.ipynb",
             "SuperPixelSaliency.ipynb",
             "SwappableImplementations.ipynb",
-            # Model comparison example is excluded due to computational complexity
-            # "VIAME_OcclusionSaliency.ipynb",
+            # VIAME notebook takes longer than 60 minutes to execute (timeout limit)
```
bjrichardwebster: I'm fine changing this but I don't think the timeout is 60 minutes right now. If we go with this comment, we should add a timeout to the notebook jobs (potentially we could set a global 60 min timeout for all of our workflows).

.github/workflows/.github-test.yml
bjrichardwebster: Given the number of environments that we now run, I think we should add concurency groups to force single job runs at a time. Potentially --- though open to suggestion --- group all notebooks, group each python version so each version is running one at a time, and the rest can probably run by itself. Thoughts?

bjrichardwebster: Potentially one gate per python group?

bjrichardwebster:
Do we know why the code coverage is dropping so much, is it some of the new changes and we're just working on it still as we harden the package?

bjrichardwebster (changes requested):

